PR #1, opened 2021-04-27
Move the graphql VS Code extension from apollo-tooling into this repo

hwillson:
The [`apollo-tooling`](https://github.com/apollographql/apollo-tooling) repo is going to be archived in the not too distant future, which means we need to find a new home for the parts of that repo we want to continue to work on. The Apollo GraphQL VS Code extension is one of those parts.

This PR migrates over a copy of the graphql VS Code extension codebase, keeping as much of the original codebase in tact. Some adjustments have been made to make sure the extension can be built and used from this new location, but the work in this PR should be considered as a stepping stone (ie. there are many future changes that can and will be made to reduce the bit rot of this project).

A few important things to call out:

- I've tested the changes in this PR via `npm run package-extension` followed by installing the generated .vsix directly in VS Code. Everything appears to be working properly.
- I've renamed the project from `vscode-apollo` to `vscode-graphql`, as we now have multiple Apollo VS Code extensions (graphql, workbench, theme). I don't think the new name will impact existing users when deployed to the MS marketplace, since the detailed name and description of the extension are staying the same, but just a heads up incase. If it looks like this change will lead to the extension showing up as a new extension in the marketplace for some reasons, instead of an update to the existing extension, we can change it back.
- This extension was dependent on several packages stored in the `apollo-tooling` repo (e.g. `apollo-env`, `apollo-language-server`, `apollo-tools`). At some point in the future, common utility packages like `apollo-tools` will likely be [moved out into their own repo](https://github.com/apollographql/federation/issues/685), but for now we need to make sure the VS Code extension can be developed separately from `apollo-tooling`. To help with this, copies of the dependent code this extension needs have been moved into the `src` directory (e.g. `src/env`, `src/language-server`, `src/tools`). This can be adjusted as needed in the future.

**TODO:**

- [ ] Figure out the testing infra; it isn't actually working in the `apollo-tooling` repo
- [ ] Review the marketplace push config, to make sure we're ready to publish from this new repo
- [ ] More extensive testing to make sure all changes are working properly (e.g. make sure performance metrics from Studio are working properly)


hwillson:
👋 @cheapsteak - sorry for the delay! 😬 Would you prefer that I merge this PR as is, then do the handoff to your team for the next steps? This gets things into a working state (I've been testing with a local build of the extension and everything seems to work), but I haven't done extensive testing. Maybe now is a good time for the handoff regardless, as I'm proving to be a blocker time wise here, and I'm sure your team will want to slice and dice further anyways. Let me know - thanks!

cheapsteak:
@hwillson Yes please that'd be great!  
Would be happy to pick up the repo from here, starting with testing as first steps  
Thanks incredibly for the work already done 🙏

hwillson:
Sounds great @cheapsteak - done, and over to you. Thanks very much! 🙇‍♂️

PR #4, opened 2021-07-22
fix: support `type: module` with `.cjs`file extension

alexanderniebuhr:
ported over from
https://github.com/apollographql/apollo-tooling/pull/2381/


alexanderniebuhr:
i do understand if you do not want to add support for `esm` yet, but this commit just allows `cjs` support with an additional file extension. This is a major break for us, why we have to run our custom builded version of this extension

/cc @jgzuke @cy

cy:
Looks reasonable, thank you @alexanderniebuhr

alexanderniebuhr:
sorry for missing prettier run. I guess I missed it, because I ported this PR.

PR #6, opened 2021-07-27
Get most tests working

jgzuke:
A few changes here
- Most of the tests in this repo are now in a working state
- A bunch of typescript errors are now fixed, so typescript should now run happily as well
- Added in a lightweight prettier config, matching the existing code style

`statusBar` tests require `vscode` to run, which was the reason behind the old `testRunner`, however I think this is no longer needed. Instead of trying to get the old testRunner working, will try to move to the [more modern way](https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode) of writing vscode tests. Going to do that in a separate PR.

Easiest reviewed by commit, individual commit messages should explain each change.


src/language-server/config/loadConfig.ts
```diff
@@ -72,7 +72,7 @@ export async function loadConfig({
   requireConfig = false,
   name,
   type
-}: LoadConfigSettings) {
+}: LoadConfigSettings): Promise<ApolloConfig | undefined> {
```
cy: how about returning `ApolloConfig | null ` instead of `undefined` so that it's easier to tell that an empty value is intentionally being returned vs an uninitialized value

src/tools/__tests__/snapshotSerializers/astSerializer.ts
```diff
@@ -1,24 +1,26 @@
...
   },
 
   serialize(
     value: ASTNode,
-    config: Config,
-    indentation: string,
-    depth: number,
-    refs: Refs,
-    printer: Printer
+    _config,
```
cy: this is prob a dumb question, but if the params are unused, why not delete them instead of per-fixing them with `_`

package.json
```diff
@@ -70,6 +70,7 @@
     "jest": "^26.6.3",
     "jest-environment-node": "^26.6.2",
     "memfs": "^3.2.2",
+    "prettier": "^2.3.2",
```
cy, line 73: 🥳

src/__tests__/statusBar.test.ts
```diff
@@ -1,6 +1,8 @@
 import StatusBar from "../statusBar";
 
-describe("statusBar", () => {
+// TODO (jgzuke) disable this until we migrate away from `vscode`.
+// https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode
+describe.skip("statusBar", () => {
```
cy, line 5: even though the test is skipped, `yarn test` still fails on the import `> 1 | import { window, StatusBarAlignment } from "vscode";` 🤔 maybe this file could be commented out instead of skipped?

cy (approved):
left some comments but LGTM otherwise

src/tools/__tests__/snapshotSerializers/astSerializer.ts
jgzuke: They don't need to be here (besides `_config`). I had originally left them before we actually had type safety as a hint, but now that we're using the `Plugin` type over asserting it, I dont feel strongly about keeping them 🤷‍♂️ Will remove

src/language-server/config/loadConfig.ts
jgzuke: Yep 💯

PR #8, opened 2021-07-28
fix: Don't load localSchemaFile as part of project code

jgzuke:
This should fix what I believe is the most common case for hitting https://github.com/apollographql/apollo-tooling/issues/801 (VSCode Plugin: Loading Schema Twice).

When using a local schema file to power the VSCode extension, if the schema file is in the `includes` (which it probably is by default), we will read it as part of the project code and try to extend the existing schema. Since this file has already been included as the base, the extension will fail and show an error.

This change adds the local schema file(s) to the `excludes` list for the file set, which I think this makes sense overall. Excluding this from the project does mean that it wont get included for providers when editing this file, however in practice I dont think this worked anyways, and I dont see a strong use case for supporting editing a local schema file anyways assuming it did work.

First three commits are refactors to make the intent of the final commit more clear.

### DIFF
--- a/src/language-server/config/config.ts
+++ b/src/language-server/config/config.ts
@@ -208,8 +208,12 @@ export class ApolloConfig {
 
 export class ClientConfig extends ApolloConfig {
   public client!: ClientConfigFormat;
+  public isClient!: true;
+  public isService!: false;
 }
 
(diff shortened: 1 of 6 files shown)


cy (approved):
makes sense, and still runs so :shipit:

PR #10, opened 2021-08-02
cleanup: Switch fully to npm from from npm/yarn hybrid

jgzuke:
This subproject seemed to be some sort of yarn/npm hybrid. Let's fully remove the yarn references. I think this makes sense given that the people on the team know npm better, and its probably better supported for `vsce` 🤷‍♂️

This removes some manual `node_modules` management, regenerates package-lock, and fixes some scripts/types. More context in commit descriptions.

This increases the size of the produced vsix file. I will follow this up with a switch to properly bundling before packaging so this shouldn't matter.


cy (approved):
Great! Could you also include an `.nvmrc` with expected node ver, cuz when I first tried this branch I was unwitting using system default node ver of 12.16.1 and it generated a package-lock.json diff. Switching to using `lts/erbium` from `studio-ui'`s `.nvmrc` caused no diff, so I'm assuming you're using `lts/erbium`?

jgzuke:
Good call 👍 Will test with something more modern later and bump, at least to `fermium`.

Going to merge over ghost check

PR #11, opened 2021-08-02
cleanup: Set up circle for project

jgzuke:
Adds a circle config with checks for ’lint’/‘typecheck’/‘test-unit’. Two of these will output reports. Typescript doesn’t support xml output, but that doesn’t seem that helpful anyways.

This also switches to running prettier through eslint. This lets us have one unified entry point, but also lets us generate xml output. More context on that commit description

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,70 @@
+version: 2.1
+
+executors:
+  node:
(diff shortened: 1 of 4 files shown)


cy (approved):
✅

PR #20, opened 2021-09-09
Add 'Run in explorer' to operations

jgzuke:
First couple commits are cleanup, last ones have the functional changes.

The idea here is to allow users to open an operation without having to traverse their fragment tree to collect the full runnable operation. This will load Explorer in a Studio registered graph if you specified your `client.service` in your config as a `graph@variant` string, otherwise it will load Eplorer with your endpoint from your `config.service` if you are using introspection.

Would love to have clicking the button directly open the link, but unfortunately VSCode [does not allow click events on gutter items](https://github.com/Microsoft/vscode/issues/5455).

https://share.getcloudapp.com/o0uP94lx

┆Issue is synchronized with this [Jira Task](https://apollographql.atlassian.net/browse/NEBULA-438) by [Unito](https://www.unito.io)


src/extension.ts
```diff
@@ -206,36 +207,60 @@ export function activate(context: ExtensionContext) {
...
+              window.activeTextEditor!.document.uri.toString()
           )
-          .map((dec) => {
+          .map((decoration): DecorationOptions => {
+            const onDiskPath = Uri.file(
+              join(context.extensionPath, "src", "iconRun.svg")
+            );
+
+            const hoverMessage =
+              decoration.hoverMessage === undefined
```
cy, line 227: nit: is this equivalent to `decoration.hoverMessage && new MarkdownString(decoration.hoverMessage)`

cy (approved):
This is super cool! 🤩

src/extension.ts
jgzuke: The empty string being falsy makes me prefer explicit `undefined` checks 🤷‍♂️ I would rather have empty string show an empty hover.

PR #24, opened 2021-09-18
fix: Misc build fixes for publish

jgzuke:
- When preparing the extension for publish we need dev deps to to the build
- Icon should be moved outside `/src` so it has the same absolute path in published and dev builds
- Switch some error returns to be explicit any for now


mayakoneval (approved):
👍

PR #27, opened 2021-09-25
Update Run in Explorer button icon

jglovier:
This PR updates the Run in Explorer button icon to match the Run in Explorer button in Studio.

| Before | After |
|---|---|
|image|image|


jglovier:
🤔 ...for some reason it appears the new icon is pushing the text down. In the before, the icon, line number, and text row all appear aligned-middle. In the after image, the text row appears aligned-baseline. I'd like them to be aligned-middle, but not sure how to adjust this.

jgzuke:
@jglovier  https://github.com/apollographql/vscode-graphql/pull/32 should fix the icon pushing down the text (and the pushing to the right as well)

jgzuke:
@cy mind doing a quick double check here?

cy:
Screen Shot 2021-10-12 at 4 09 33 PM
The button overlaps the first character of `query` for me - is it just my resolution? Zooming in and out still has the same issue.

Screen Shot 2021-10-12 at 4 12 15 PM
Changing the default theme has the overlap too

jgzuke:
@cy only added the negative margin when text was indented, does that seem like a decent solution?

cy (approved):
yeah. Looks good now!

jglovier:
Is the triangle balanced a little low for anyone else, or is it just me?

image

(also not sure how it got filled in...if we want to be consistent with Studio it needs to just be the outline)

Issue #18, opened 2021-09-08
Falsy "️️There are multiple definitions for the operation" triggered


yhaskell:
I see a lot of failures in the Apollo GraphQL Output for vscode, that looks like this:

    [Error - 18:28:53] Notification handler 'workspace/didChangeWatchedFiles' failed with message: ️️There are multiple definitions for the `OperationName` operation. Please rename or remove all operations with the duplicated name before continuing.When I search for this operation, I don't find it anymore. 
Restarting vscode helps, but not for a long time.

Can we please have in this error actual places where it finds the problems?
That would help to understand the issue more, and actually help immensely when this problem actually occur.

Thank you!

┆Issue is synchronized with this [Jira Task](https://apollographql.atlassian.net/browse/NEBULA-433) by [Unito](https://www.unito.io)

anders-photowall:
I have the same or similar issue. The issue for me is that as soon as you open a non-file version of the file containing the definition (i.e. working-tree version or really any diff) the apollo extension reads that and treats it as a duplicate definition. Before finding this issue I found this pull request on the old extension repo which actually fixes the issue but it was never merged: 
https://github.com/apollographql/apollo-tooling/pull/1894

jgzuke:
This is my next priority to get fixed! Hoping to get to it soon but unfortunately haven't had time

memark:
Thank you @jgzuke, looking forward to it!

jgzuke:
This should be fixed in the latest `1.19.7` release 😄

PR #28, opened 2021-10-06
fix: Ignore edits/creations for non file documents

### LINKED ISSUE #1789: Apollo GraphQL vscode extension will panic when view git diff
**Intended outcome:**

no panic when view git diff

**Actual outcome:**
panic when view git diff 

**How to reproduce the issue:**

add the blow content to `a.ts` and commit it to git , then use editor git diff view diffrence, the extension panic and throw error: ` ️️There are multiple definitions for the GetGithubConfig operation. All operations in a project must have unique names. If generating types, only the types for the first definition found will be generated.`
```ts
import { gql } from '@apollo/client'
let a = gql`
query something {
hello world
}
`
```

**Versions**

1.14.0

jgzuke:
This ignores document edits/creations that are coming from non file sources (eg merge view documents).

Previously if you opened the diff viewer (and possibly other ephemeral documents) the extension would see multiple definitions with the same name and crash.

Adapted from: https://github.com/apollographql/apollo-tooling/pull/1894
Fixes: https://github.com/apollographql/apollo-tooling/issues/1789, https://github.com/apollographql/apollo-tooling/issues/1875, part of https://github.com/apollographql/vscode-graphql/issues/18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Next
 
 - Specify files with conflicting documents for the 'There are multiple definitions' message. [#29](https://github.com/apollographql/vscode-graphql/pull/29)
+- Dont watch non file documents. Diff view will no longer crash extension. [#28](https://github.com/apollographql/vscode-graphql/pull/28)
 
 ### 1.19.6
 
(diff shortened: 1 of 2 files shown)


cy (approved):
✨ works

PR #29, opened 2021-10-06
fix: Specify file names for conflicting operation definition message

jgzuke:
Improve 'Conflicting definitions' message to specify files that are causing the issue.

Fixes: Part of https://github.com/apollographql/vscode-graphql/issues/18


cy (approved):
✔️

PR #32, opened 2021-10-08
Dont push down the line text with larger icon

jgzuke:
Looks like we cant use the full icon without pushing down the text on the line, so the icon here is just the inner ‘>’, with the blue background being added with css. This *should* look about the same as what you had in your PR, besides the while fill you mentioned.

Theres a bit of associated message cleanup here as well that should be non functional but made sense to add with the icon update.

Screen Shot 2021-10-08 at 3 33 17 PM


jglovier (approved):
:bow: thank youuu!!

PR #36, opened 2021-10-20
fix: Fix syntax highlighting for implements with directive

jgzuke:
```graphql
type Product implements ProductItf @key(fields: "id") {
  …
}
```
would previously not follow the pattern `"begin": "\\s*\\b(implements)\\b\\s*", "end": "\\s*(?={)",` because the `implements ProductItf` was followed by an ‘@’, not a ‘{’ (which is perfectly valid). This meant this was not getting recognized as an interface.

You might be asking, why do we have our own full grammar defined here when it probably isn't up to date with the spec, there must be a better way 💭 Agree, but let's fix this issue for now 😛

A few other random small cleanups here as well. The first commit is the functional change.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -8,6 +8,7 @@ src
 .gitignore
 .nvmrc
 .prettierrc
+.gitattributes
 codegen.yml
 graphql.configuration.json
 jest.config.ts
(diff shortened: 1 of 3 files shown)


cy (approved):
🔥

Issue #37, opened 2021-10-21
Is there a way to disable the "run in studio" button through some vscode setting?


benatshippabo:
Is there a way to disable the "run in studio" button through a vscode setting? I find the increased indentation disorienting.

![image](https://user-images.githubusercontent.com/66273043/138192832-ad07bdca-b6f9-4957-afc5-ea25d2638f3f.png)

┆Issue is synchronized with this [Jira Task](https://apollographql.atlassian.net/browse/NEBULA-551) by [Unito](https://www.unito.io)

ivnnv:
+100
To this should be an opt-in (off by default) option because having a config file doesn't necessarily means we have access or want to use studio.

talon-i-am:
+1

whengely:
Please add a way to disable this.  Not only does it clutter my screen, it also pushes text further to the right so it doesn't line up.

jgzuke:
Hey y'all, Sorry for the wait! We ended up moving the button to the end of the line after the text over in https://github.com/apollographql/vscode-graphql/pull/42, we're hoping that this solves:
- Increased indentation / pushing the text to the right
- Extra clutter of the screen
- Allows people who don't want to use Sandbox to ignore the button

I'm hesitant to add a config given that the Apollo config in it's current state is honestly pretty confusing (which I hope we can fix!). I'm hoping that this being moved out of the way will alleviate the need for people to turn it off, while still allowing those that have started using it to continue, and not needing to add any complexity to the config.

Thank you for raising this issue! Honestly it fell off my radar for a while, sorry it took so long to get to 🙏

Going to close for now, if this is still an issue after this change let us know!

yusufumac:
Jason, it is a clutter if some people don't use it and is not easy to ignore a big blue button on every query. As far as I know, Apollo VS code extension doesn't have many configs. Adding a config option to hide these button while keeping the default as shown won't affect the current users but it gives the ability to hide it for people who is never going to use it.

ivnnv:
Man... this is such a cheap solution... 😄 

![button](https://user-images.githubusercontent.com/23552631/141266170-65e1f5e1-7dcd-446c-af38-f37d5503a313.png)

benatshippabo:
@jgzuke,  @yusufumac is right, there is currently only one other vscode configuration setting:

https://github.com/apollographql/vscode-graphql/blob/89d014a1069845b203ad36a2962b2b94822f8292/package.json#L104-L120

[...]

PR #42, opened 2021-11-10
Move 'Run in Studio' icon to the right of the operation signature

jgzuke:
Moves the icon to the right of the operation signature, this will stop it from pushing around text and keep it more out of the way.

Screen Shot 2021-11-09 at 11 20 00 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+### vNext
+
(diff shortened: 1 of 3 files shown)


papollosc:
@jgzuke I thought we were going to put it at the very end of the line?

cy (approved):
LGTM

jgzuke:
> I thought we were going to put it at the very end of the line?

Resolved in person, not going to make this change, there are some limits on what we can do here in an extension

benatshippabo:
Could we also have a vscode setting to hide this button as well? 🙏  #37

PR #50, opened 2021-11-18
feature: implement elixir support based only on sigils

pmargreff:
Hey, first of all, thank you for the tooling. Apollo has been an excellent tool in the GraphQL ecosystem 🙇🏾 . In this PR, I'm adding an alternative syntax for using `vscode-graphql` with Elixir, some context why: 

#### Context
The original implementation requires the GraphQL block to be wrapped in a `gql` function. That forces users to define an empty `gql` function if the devs want to use `vscode-graphql` features even with no utility. I took a look at the original PR (apollographql/apollo-tooling#1304), and it was intentional. It seems like the original author had this pattern already defined in his codebase.

#### Solution
One option is to use [sigils](https://elixir-lang.org/getting-started/sigils.html#strings-char-lists-and-word-lists-sigils) for improving it. Sigils also allow heredocs, and heredocs is how `vscode-graphql` supports ruby (apollographql/apollo-tooling#1304).

By default, elixir allows users to use `~s"""...` for strings without interpolation and `~S"""...` for cases where string interpolation or escaping is needed. Elixir is a very extensible language, allowing users to [write their own sigils](https://elixir-lang.org/getting-started/sigils.html#custom-sigils). This implementation makes `vscode-graphql` shine when users have defined sigils in a heredoc style for graphql operations.

Once users may be using the original format, I kept that working together with the new syntax support. I also added basic tests to ensure things are still working for the version based on function and the version based on sigils.

### Screenshots
vscode-elixir-with-sigils

┆Issue is synchronized with this [Jira Task](https://apollographql.atlassian.net/browse/NEBULA-647) by [Unito](https://www.unito.io)

### DIFF
--- a/syntaxes/graphql.ex.json
+++ b/syntaxes/graphql.ex.json
@@ -35,6 +35,17 @@
           ]
         }
       ]
+    },
+    {
+      "name": "taggedTemplates",
+      "contentName": "meta.embedded.block.graphql",
+      "begin": "\\~(gql|graphql|GQL|GRAPHQL)(\"\"\")",
+      "end": "((\\2))",
+      "patterns": [
+        {
+          "include": "source.graphql"
+        }
+      ]
     }
   ],
   "scopeName": "inline.graphql.elixir"
--- a/src/language-server/__tests__/document.test.ts
+++ b/src/language-server/__tests__/document.test.ts
@@ -157,4 +157,68 @@ describe("extractGraphQLDocuments", () => {
       expect(documents?.[0].ast?.definitions.length).toBe(1);
     });
   });
+
+  describe("extracting documents from Elixir extension nodes", () => {
+    const mockTextDocument = (text: string): TextDocument => ({
+      getText: jest.fn().mockReturnValue(text),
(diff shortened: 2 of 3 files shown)


syntaxes/graphql.ex.json
```diff
@@ -35,6 +35,17 @@
           ]
         }
       ]
+    },
+    {
+      "name": "taggedTemplates",
+      "contentName": "meta.embedded.block.graphql",
+      "begin": "\\~(gql|graphql|GQL|GRAPHQL)(\"\"\")",
```
pmargreff, line 42: I have added both high and low cases because each definition has different meanings. About adding `gql` and `graphql` I have seen both implementations in different code bases over time. The intention is to make everyone happy by default andavoidingd code rewrites if users want to try `vscode-graphql` extension.

pmargreff:
Putting it on hold, seems like Elixir does not allow users to define a sigil with multiple chars. I will try to make that happen in Elixir lang first then I will be back to this PR. 😢

Issue #64, opened 2022-04-01
`TypeError: Cannot read properties of undefined (reading 'name')`


lhrnd0:
Receiving this after the latest VSCode update. Anyone with the same error?

```
/.vscode/extensions/apollographql.vscode-apollo-1.19.9/lib/language-server/errors/validation.js:103
                selectsClientFieldSet = Boolean(clientFields && clientFields.includes(fieldDef.name));
                                                                                               ^

TypeError: Cannot read properties of undefined (reading 'name')
```

┆Issue is synchronized with this [Jira Task](https://apollographql.atlassian.net/browse/NEBULA-1075) by [Unito](https://www.unito.io)

NataliaTepluhina:
Same error after the update

PR #73, opened 2022-05-24
Fix for #64: return before error on missing fields

rachsmithcodes:
I am getting the error described in https://github.com/apollographql/vscode-graphql/issues/64.

As far as I can tell, it is happening in when a GraphQL document includes a field that doesn't exist on the server schema. In this scenario the validation should add an error like this:

CleanShot 2022-05-24 at 13 07 29

But when it gets run though the `NoMissingClientDirectives` rule, `context.getFieldDef()` returns `undefined`, causing the extension to error on line 103 of `validation.js`. 

Adding an early return if the `fieldDef` doesn't exist prevents the error.

┆Issue is synchronized with this [Jira Task](https://apollographql.atlassian.net/browse/NEBULA-1294) by [Unito](https://www.unito.io)

### DIFF
--- a/src/language-server/errors/validation.ts
+++ b/src/language-server/errors/validation.ts
@@ -175,7 +175,7 @@ export function NoMissingClientDirectives(context: ValidationContext) {
     const fieldDef = context.getFieldDef();
 
     // if we don't have a type to check then we can early return
-    if (!parentType) return;
+    if (!parentType || fieldDef) return;
 
     // here we collect all of the fields on a type that are marked "local"
     const clientFields =
--- a/src/language-server/errors/validation.ts
+++ b/src/language-server/errors/validation.ts
@@ -175,7 +175,7 @@ export function NoMissingClientDirectives(context: ValidationContext) {
     const fieldDef = context.getFieldDef();
 
     // if we don't have a type to check then we can early return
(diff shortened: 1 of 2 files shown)


src/language-server/errors/validation.ts
```diff
@@ -175,7 +175,7 @@ export function NoMissingClientDirectives(context: ValidationContext) {
     const fieldDef = context.getFieldDef();
 
     // if we don't have a type to check then we can early return
-    if (!parentType) return;
+    if (!parentType || fieldDef) return;
```
pauljohnberry: Perhaps reverse the check on fieldDef here? `!fieldDef`

alessbell: Agree it looks like we need to flip this condition, went ahead and made that change in https://github.com/apollographql/vscode-graphql/pull/73/commits/9d95cacf0e70b7bbf3f43678c60736072ebd49dc

alessbell (approved):
Thanks @rachsmithcodes!

alessbell:
Fixed broken test in a follow-on commit: https://github.com/apollographql/vscode-graphql/commit/bf800116486b5b037f8888f58b918f40dd895d37

PR #106, opened 2023-08-11
Update README.md: set correct url to deploy a schema to the registry

marfrede:


bignimbus (approved):
Thanks for catching this!

PR #148, opened 2024-07-18
modernization - get `graphql` to 16, update codegen, drop packages

phryneas:


src/language-server/diagnostics.ts
```diff
@@ -2,10 +2,19 @@ import {
...
   GraphQLError,
   FragmentDefinitionNode,
-  findDeprecatedUsages,
   isExecutableDefinitionNode,
+  DocumentNode,
+  validate,
+  NoDeprecatedCustomRule,
 } from "graphql";
 
+function findDeprecatedUsages(
```
phryneas, line 11: This function has been removed from `graphql`, so I inlined it.

src/language-server/engine/GraphQLDataSource.ts
```diff
@@ -1,15 +1,37 @@
...
-  ForbiddenError,
-} from "apollo-server-errors";
+import { ApolloLink, execute, GraphQLRequest } from "@apollo/client/link/core";
+import { toPromise as makePromise } from "@apollo/client/link/utils";
+import { onError } from "@apollo/client/link/error";
+import { setContext } from "@apollo/client/link/context";
+import { createHttpLink } from "@apollo/client/link/http";
 import to from "await-to-js";
 import { GraphQLError } from "graphql";
+class ApolloError extends Error {
```
phryneas, line 9: This whole file is already an inlined & modified version of https://www.npmjs.com/package/apollo-datasource-graphql (not an actual apollo package), and the `ApolloError` from `apollo-server-errors` was not compatible with `graphql` 16, so I replaced these three classes with inline kinda-similar versions.

In a future PR, I'll probably replace this whole file with `@apollo/client`, since all it is used for is make three different queries.

src/language-server/providers/schema/file.ts
```diff
@@ -3,22 +3,17 @@ import {
...
-  visit,
+  buildASTSchema,
 } from "graphql";
+import { mergeTypeDefs } from "@graphql-tools/merge";
 import { readFileSync } from "fs";
 import { extname, resolve } from "path";
 import { GraphQLSchemaProvider, SchemaChangeUnsubscribeHandler } from "./base";
 import { NotificationHandler } from "vscode-languageserver/node";
 import { Debug } from "../../utilities";
-import { buildSchemaFromSDL } from "apollo-graphql";
```
phryneas, line 16: `apollo-graphql` still depends on `graphql` 15, and we only use this one function (source: https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo-graphql/src/schema/buildSchemaFromSDL.ts#L87)

I *hope* my replacement code here is equivalent, but this is what I'm most unsure about in this PR.

src/language-server/utilities/graphql.ts
```diff
@@ -307,16 +308,16 @@ const typenameField = {
   name: { kind: Kind.NAME, value: "__typename" },
 };
 
-export function withTypenameFieldAddedWhereNeeded(ast: ASTNode) {
+export function withTypenameFieldAddedWhereNeeded(ast: DocumentNode) {
   return visit(ast, {
-    enter: {
-      SelectionSet(node: SelectionSetNode) {
+    SelectionSet: {
+      enter(node: SelectionSetNode) {
```
phryneas, line 314: This 4th syntax was removed from `visit` in `graphql` 16.

tsconfig.json
```diff
@@ -19,8 +19,7 @@
     "forceConsistentCasingInFileNames": true,
     "lib": ["es2020", "esnext.asynciterable"],
     "types": ["node", "jest"],
-    "baseUrl": ".",
-    "skipLibCheck": true
```
phryneas, line 23: 🎉

phryneas (commented):
Gonna do a round of manual "trying out" if things still work tomorrow, and then it's essentially ready for review.

package.json
```diff
@@ -30,24 +30,20 @@
     "vscode": "^1.90.0"
   },
   "dependencies": {
-    "@apollo/federation": "^0.23.1",
+    "@apollo/client": "^3.10.8",
```
jerelmiller, line 33:
```suggestion
    "@apollo/client": "^3.11.1",
```
?

src/language-server/engine/GraphQLDataSource.ts
```diff
@@ -1,15 +1,37 @@
...
+  public extensions?: Record<string, any>;
+
+  constructor(message: string, code?: string) {
+    super(message);
+    if (code) {
+      this.extensions = { code };
+    }
+    Object.defineProperty(this, "name", { value: "ApolloError" });
```
jerelmiller, line 17: For my own learning/understanding, why use `defineProperty` here to set `name` rather than just `this.name = "ApolloError"`?

jerelmiller (approved):
Looks great! Nice to get everything upgraded

package.json
phryneas: I'll let dependabot handle that, otherwise I have to rebase 4 PRs 😅

src/language-server/engine/GraphQLDataSource.ts
phryneas: That's just copy-pasted from the upstream implementation, where I just removed 2/3 of the code - I didn't think about it further as I plan to remove all of this soon anyways :)

PR #151, opened 2024-07-25
introduce type-safe message passing

phryneas:
Right now, type safety is a little bit "all over the place" - this fixes that.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,8 +15,13 @@ import {
 } from "vscode";
 import StatusBar from "./statusBar";
 import { getLanguageServerClient } from "./languageServerClient";
-import { NotificationType } from "vscode-languageclient/node";
-import type { EngineDecoration, LanguageClient } from "./messages";
+import { LanguageClient, NotificationType } from "vscode-languageclient/node";
+import {
+  type EngineDecoration,
+  LanguageServerCommands as LSCommands,
+  LanguageServerNotifications as LSNotifications,
+  LanguageServerRequests as LSRequests,
+} from "./messages";
 import {
   printNoFileOpenMessage,
   printStatsToClientOutputChannel,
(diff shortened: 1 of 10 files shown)


src/utils.ts
```diff
@@ -1,4 +1,5 @@
-import type { LanguageClient, ProjectStats } from "./messages";
+import { LanguageClient } from "vscode-languageclient/node";
```
jerelmiller, line 1: Should this remain an `import type`?

src/language-server/utilities/debug.ts
```diff
@@ -1,4 +1,5 @@
-import { Connection } from "src/messages";
+import { LanguageServerNotifications as Notifications } from "../../messages";
+import { Connection } from "vscode-languageserver/node";
```
jerelmiller, line 2:
```suggestion
import type { Connection } from "vscode-languageserver/node";
```

Not sure if we are planning to use `import type` where possible in this project or not, but figured I'd spot some of the places that look like type imports 🙂

src/extension.ts
```diff
@@ -15,8 +15,13 @@ import {
...
 import StatusBar from "./statusBar";
 import { getLanguageServerClient } from "./languageServerClient";
-import { NotificationType } from "vscode-languageclient/node";
-import type { EngineDecoration, LanguageClient } from "./messages";
+import { LanguageClient, NotificationType } from "vscode-languageclient/node";
+import {
+  type EngineDecoration,
+  LanguageServerCommands as LSCommands,
+  LanguageServerNotifications as LSNotifications,
+  LanguageServerRequests as LSRequests,
```
jerelmiller, line 23: I see we use `Commands`, `Notifications`, etc for these aliases in other files. Should this be consistent with those?

jerelmiller (approved):
Looks great! Feel free to take or leave my comments. Absolutely nothing blocking there. Always great to see more type safety!

src/extension.ts
phryneas: The other files are inside of the Language Server, this is the only file "outside" the language server, so I explicitly added it for stated perspective here.

src/language-server/utilities/debug.ts
phryneas: Tbh., this will keep happening a lot. I'll do a cleanup swipe over all of these after we have a few PRs in :)

src/extension.ts
jerelmiller: Makes sense!

PR #153, opened 2024-07-26
add `@defer` and `@nonreactive`, remove old config options

phryneas:
This adds our "new" directives to be available in client schemas.

It also removes some old config options that were only used in unreferenced code - in the past, that code was referenced from the CLI that the VSCode extension was originally based on.


jerelmiller (approved):

apollo-bot2:
## Detected SAST Vulnerabilities

PR #159, opened 2024-07-30
Fix a bug where unrelated project configurations could be merged into each other.

phryneas:


src/language-server/config/loadConfig.ts
```diff
@@ -206,11 +206,13 @@ export async function loadConfig({
   // selectively apply defaults when loading the config
   // this is just the includes/excludes defaults.
   // These need to go on _all_ configs. That's why this is last.
-  if (config.client) config = merge({ client: DefaultClientConfig }, config);
```
phryneas, line 209: Lodash `merge` deep-merges into the first argument, so this would actually modify `DefaultClientConfig`, which would then provide defaults to the next config file that would be loaded.

src/language-server/workspace.ts
```diff
@@ -196,6 +197,9 @@ export class GraphQLWorkspace {
     }
     // If project exists, update the config
     if (project && config) {
+      if (equal(project.config.rawConfig, config.rawConfig)) {
+        return;
+      }
```
phryneas, line 202: There are situations where a lot of "file changes" events are emitted (not sure why yet) - but we only want to reload services if the config actually changed, as reloading services can e.g. cause a network request.

If we can avoid it, we don't want to DDOS ourselves 😅

jerelmiller (approved):
Nice catch on this one!

PR #161, opened 2024-08-02
Modernization 3

phryneas:


src/language-server/project/client.ts
```diff
@@ -28,7 +28,7 @@ import {
   DiagnosticSeverity,
 } from "vscode-languageserver/node";
 import LZString from "lz-string";
-import { stringifyUrl } from "query-string";
```
phryneas, line 31: New versions of `"query-string"` are ESM only and I really don't want to touch the build process of the Extension right now.

src/language-server/project/client.ts
```diff
@@ -412,7 +412,7 @@ export class GraphQLClientProject extends GraphQLProject {
                     this.frontendUrlRoot ?? "https://studio.apollographql.com";
 
                   const variant = this.config.variant;
-                  const graphId = this.config.graph;
+                  const graphId = this.studioGraphId;
```
phryneas, line 415: If I have a config file like
```
module.exports = {
  client: {
    service: {
      name: "localSchema",
      localSchemaFile: "./starwarsSchema.graphql",
    },
  },
};
```
`graphId` would be `localSchema` - of course, linking to studio for that is kinda useless, but that's exactly what the extension did until now.

.changeset/five-houses-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"vscode-apollo": patch
+---
+
+Only show "Run in Studio" gutter action for local graphs if an endpoint is configured.
```
phryneas, line 5: I'm not 100% sure.. it seems like a good idea, but might someone get use from a link to explorer that doesn't fill in the endpoint?
My gut says removing this is the right thing, but I'd love a second opinion.

jerelmiller (approved):
💯

PR #163, opened 2024-08-07
Remove Service projects, centralized configuration handling

phryneas:
Please note that at this point, configuration verification errors might not be displayed due to a timing bug in the extension. I will tackle that bug in a follow-up PR, but this is already getting big, so I'll open it for review.


jest.config.ts
```diff
@@ -6,10 +6,16 @@ const config: Config.InitialOptions = {
     escapeString: true,
     printBasicPrototype: true,
   },
-  testMatch: ["**/__tests__/**/*.test.ts"],
+  testMatch: ["**/__tests__/**/*.ts"],
+  testPathIgnorePatterns: [
+    "/node_modules/",
+    "/fixtures/",
+    "/snapshotSerializers/",
+  ],
```
phryneas, line 14: We were missing out on actually running a bunch of tests that were not ending on `.test.ts`

jest.config.ts
```diff
@@ -6,10 +6,16 @@ const config: Config.InitialOptions = {
...
   transform: {
     "^.+\\.(ts)$": "ts-jest",
   },
+  prettierPath: null,
```
phryneas, line 18: Cannot update snapshots with prettier 3

src/language-server/config/config.ts
```diff
@@ -1,146 +1,240 @@
 import { dirname } from "path";
-import merge from "lodash.merge";
-import { ClientID, ServiceID, ServiceSpecifier } from "../engine";
 import { URI } from "vscode-uri";
-import { WithRequired } from "../../env";
 import { getGraphIdFromConfig, parseServiceSpecifier } from "./utils";
+import { Debug } from "../utilities";
+import z, { ZodError } from "zod";
```
phryneas, line 5: Moving everything over to zod. GitHub might collapse the file, but this is where the magic happens.

src/language-server/config/__tests__/config.ts
```diff
@@ -1,128 +1,54 @@
...
       // can be either /test/dir/name or \\test\\dir\\name depending on platform
       // this difference is fine :)
-      expect(config.configDirURI?.fsPath).toMatch(
+      expect(config?.configDirURI?.fsPath).toMatch(
         /\/test\/dir\/name|\\test\\dir\\name/,
       );
     });
   });
 
-  describe("projects", () => {
```
phryneas, line 32: This PR is dropping support for multiple project types in the same folder, which is what's behind this `projects` property.

src/language-server/config/__tests__/config.ts
```diff
@@ -1,128 +1,54 @@
...
-        client: { service: "hai@master", ...DefaultConfigBase },
+      const config = parseApolloConfig({
+        client: { service: "hai@master" },
       });
-      config.variant = "new";
-      expect(config.variant).toEqual("new");
-    });
-  });
-
-  describe("setDefaults", () => {
```
phryneas, line 90: This was used nowhere, so I removed it.

src/language-server/config/__tests__/loadConfig.ts
```diff
@@ -69,7 +80,7 @@ describe("loadConfig", () => {
   describe("finding files", () => {
     it("loads with client defaults from different dir", async () => {
       writeFilesToDir(dir, {
-        "my.config.js": `
+        "apollo.config.js": `
```
phryneas, line 83: No more support for custom file names, it was not used by the extension anyways.

src/language-server/config/__tests__/loadConfig.ts
```diff
@@ -160,25 +156,23 @@ describe("loadConfig", () => {
 
     it("loads config from a ts file", async () => {
       writeFilesToDir(dir, {
-        "apollo.config.ts": `module.exports = {"client": {"service": "hello"}`,
+        "apollo.config.ts": `export default {"client": {"service": "hello"} }`,
```
phryneas, line 156: This might have changed through the `cosmiconfig` update. Probably need to highlight this in the changelog.

src/language-server/config/__tests__/loadConfig.ts
```diff
@@ -267,94 +258,53 @@ describe("loadConfig", () => {
...
 
       const config = await loadConfig({
         configPath: dirPath,
-        configFileName: "my.config.js",
       });
 
       expect(config?.client?.service).toEqual("yoshi");
     });
 
-    it("Allows setting ENGINE_API_KEY with a deprecation warning", async () => {
```
phryneas, line 311: This has been deprecated for years now. Dropped it.

src/language-server/config/config.ts
```diff
@@ -1,146 +1,240 @@
...
+});
+export type LocalServiceConfig = z.infer<typeof localServiceConfig>;
 
-export const DefaultEngineConfig = {
-  endpoint:
-    process.env.APOLLO_ENGINE_ENDPOINT ||
-    "https://graphql.api.apollographql.com/api/graphql",
-};
+const clientServiceConfig = z.preprocess(
+  (value) => value || context.getValue()?.serviceName,
```
phryneas, line 50: `context.getValue()?.serviceName` in this case is the `serviceName` parsed from the `APOLLO_KEY` value in the `.env` file in the case it's a graph-level api key.

In that case, a config like `module.exports = { client: {} }` is actually valid.

src/language-server/config/loadConfig.ts
```diff
@@ -25,55 +21,35 @@ const defaultFileNames = [
...
-  // configs as a fallback.
-  configFileName?: string;
-
-  // used when run by a `Workspace` where we _know_ a config file should be present.
-  requireConfig?: boolean;
-
-  // for CLI usage, we don't _require_ a config file for everything. This allows us to pass in
-  // options to build one at runtime
-  name?: string;
-  type?: "service" | "client";
```
phryneas, line 51: The options were adjusted to actual usage in the code base. 
Stuff always passed in isn't optional anymore, stuff that was never passed in is removed.

src/language-server/engine/index.ts
```diff
@@ -56,26 +55,15 @@ export class ApolloEngineClient {
...
-        return request;
+      createHttpLink({
+        uri: baseURL,
+        headers: {
+          ["x-api-key"]: this.engineKey,
+          ["apollo-client-name"]: this.clientIdentity.name,
+          ["apollo-client-reference-id"]: this.clientIdentity.referenceID,
+          ["apollo-client-version"]: this.clientIdentity.version,
+        },
       }),
```
phryneas, line 66: This is a slight behavioural change - while before, configuring `clientIdentiy.name` would change the whole code path towards "custom options", now default options for `referenceId` and `version` would still stay in place.

I don't think this was really intended and I'm also not sure if this was ever meant to be configurable for the VSCode extension, so I think this change is okay.

src/language-server/providers/schema/__tests__/file.ts
```diff
@@ -107,18 +107,49 @@ describe("FileSchemaProvider", () => {
...
+
+directive @external(reason: String) on OBJECT | FIELD_DEFINITION
+
+directive @tag(name: String!) repeatable on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+
+directive @extends on OBJECT | INTERFACE
+
+type Query {
+  _entities(representations: [_Any!]!): [_Entity]!
+  _service: _Service!
```
phryneas, line 124: I think this comes from a federation-related package update and only surfaced now that these tests are run.

src/language-server/config/config.ts
```diff
@@ -1,146 +1,225 @@
...
-  readonly apiKey?: string;
-}
+const clientConfig = z.object({
+  service: clientServiceConfig,
+  validationRules: z
+    .union([
+      z.array(z.custom<ValidationRule>()),
+      z.function().args(z.custom<ValidationRule>()).returns(z.boolean()),
+    ])
+    .optional(),
```
phryneas, line 62: I don't think this feature has ever been documented anywhere, but it's fully implemented, so I'm not dropping it now.

phryneas:
Please note that at this point, configuration verification errors might not be displayed due to a timing bug in the extension. I will tackle that bug in a follow-up PR, but this is already getting big, so I'll open it for review.

package.json
```diff
@@ -24,7 +24,7 @@
     "typecheck": "tsc --noEmit",
     "lint": "eslint src --ext .ts",
     "format": "prettier --write .",
-    "test": "jest",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest",
```
alessbell, line 27: TIL https://jestjs.io/docs/ecmascript-modules

src/language-server/server.ts
```diff
@@ -35,9 +35,13 @@ const workspace = new GraphQLWorkspace(
   new LanguageServerLoadingHandler(connection),
   {
     clientIdentity: {
-      name: process.env["APOLLO_CLIENT_NAME"],
-      version: process.env["APOLLO_CLIENT_VERSION"],
-      referenceID: process.env["APOLLO_CLIENT_REFERENCE_ID"],
+      name: process.env["APOLLO_CLIENT_NAME"] || "Apollo Language Server",
+      version:
```
alessbell, line 39: Is there any significance to the `146d29c0-912c-46d3-b686-920e52586be6` fallback here?

src/language-server/project/base.ts
```diff
@@ -98,9 +97,11 @@ export abstract class GraphQLProject implements GraphQLSchemaProvider {
     // if a config doesn't have a uri associated, we can assume the `rootURI` is the project's root.
     this.rootURI = config.configDirURI || configFolderURI;
 
-    const { includes, excludes } = config.isClient
+    const { includes = [], excludes = [] } = isClientConfig(config)
       ? config.client
-      : config.service;
+      : {
+          /** TODO */
```
alessbell, line 103: Should this TODO be captured in an issue? You probably have it tracked somewhere, asking just in case :)

src/language-server/engine/index.ts
alessbell: That seems reasonable - maybe add that to its own changeset on this PR so it gets called out in release notes?

phryneas: The behaviour itself was never documented to start with, so I'm secretly hoping this one dies a silent death without ever being used 😓

src/language-server/project/base.ts
phryneas: It's the main goal of my next few weeks ;)

src/language-server/server.ts
phryneas: It has just moved around - see the old [src/language-server/engine/index.ts](https://github.com/apollographql/vscode-graphql/pull/163/files#diff-d675ce2b91731f4d5c3cdb539cea2811a6c1c5bfebcc59b64e56b16db8471333)

phryneas:
Seems like I added an error in one of the late one-off commits. Fixed and CI passes now.

phryneas:
Not sure what's up with the tests after merging in main, chery-picking `--runInBand` from the other PR.

alessbell (approved):
Exciting to see all this refactoring coming together 🚀

PR #165, opened 2024-08-09
update dependencies

phryneas:
Getting ahead of Renovate here.


package.json
```diff
@@ -79,8 +78,7 @@
     "rimraf": "6.0.1",
     "ts-jest": "29.1.2",
     "ts-node": "^10.9.2",
-    "typescript": "^5.5.3",
-    "vsce": "1.96.1"
```
phryneas, line 83: `@vscode/vsce` is packaged in `HaaLeo/publish-vscode-extension@v1` so we don't need it as a dependency on our own

`vsce` itself is deprecated and we're never calling it.

alessbell (approved):

PR #166, opened 2024-08-09
move logic from `LanguageProvider` into `GraphQLClientProject`, delete `LanguageProvider`

phryneas:
No changelog since this is a pure Refactor at this point.

Essentially, most code in `LanguageProvider` was "get project, if project is not a client project, return null".
The only exception from that was "search references"-type functionality, which also worked with server projects in the past - but since server projects have been removed, that's not the case either.

With this change, a project will now fully be in charge of it's own completions, etc.


alessbell (approved):
LGTM

apollo-bot2:
## Detected SAST Vulnerabilities

PR #167, opened 2024-08-09
update README

phryneas:
This gets the README in line with the newest docs changes from https://github.com/apollographql/docs/pull/913


README.md
```diff
@@ -43,24 +44,25 @@ module.exports = {
...
 
-The service name is the id of the graph created in Apollo and can be found in the graphs dashboard of [Apollo Studio](https://studio.apollographql.com)
+The `service` name is the name of the graph you've created in [GraphOS Studio](https://studio.apollographql.com).
 
-> Important note: If the display name of the graph in Studio is changed, this value should still remain the same graph **ID**. This can be found in the url when browsing the graph in Studio.
+See [additional configuration options](#additional-apollo-config-options).
 
-<h3 id="api-key">Setting up an API key</h3>
-To authenticate with Apollo to pull down the schema, create a file next to the `apollo.config.js` called `.env`. This should be an untracked file (i.e. don't push it to GitHub). Go to the settings page of the published service and create a new API key.
+<h3 id="api-key">Setting up the `.env` file</h3>
```
jerelmiller: It looks like this isn't rendering as proper markdown since its using raw HTML.

jerelmiller (approved):
Spotted one tiny thing, but otherwise looks great! 🚢

README.md
phryneas: image

Can you elaborate? It looks fine to me: https://github.com/apollographql/vscode-graphql/blob/pr/README/README.md#setting-up-the-env-file

Generally: we can't rely on GH markdown either, it depends on how it will render out in the marketplace in the end.

phryneas:
The wordmark looks good so I'll point it to `main` where it'll be found later
image

README.md
jerelmiller: Ah sorry I should have been more specific. I was referring to the text \`env\` which is rendering the literal backticks rather than an inline code block.

phryneas: I'm blind 🤦 
Replaced it with code tags, thank you for spotting this!

PR #173, opened 2024-08-15
bugfixes, forwarding to nested LSP projects

phryneas:


package.json
```diff
@@ -43,19 +43,22 @@
     "@wry/equality": "0.5.7",
     "cosmiconfig": "9.0.0",
     "dotenv": "16.4.5",
+    "fractional-indexing": "2.1.0",
```
phryneas, line 46: using v2 of the package here because v3 is ESM only

src/language-server/project/base.ts
```diff
@@ -1,97 +1,51 @@
...
+    ? K
+    : never]: Connection[K] extends (
+    params: ServerRequestHandler<any, any, any, any> & infer P,
+    token: CancellationToken,
+  ) => any
+    ? P
+    : never;
+};
+
+export abstract class GraphQLProject {
```
phryneas, line 40: This class was split up into this `base.ts` and `internal.ts` - those "internal" changes could also have moved into the `ClientProject`, but that class is big enough as it is right now.

src/language-server/project/internal.ts
```diff
@@ -0,0 +1,346 @@
...
+    if (documents) {
+      this.documentsByFile.set(document.uri, documents);
+      this.invalidate();
+    } else {
+      this.removeGraphQLDocumentsFor(document.uri);
+    }
+    this.checkForDuplicateOperations();
+  };
+
+  checkForDuplicateOperations = throttle(
```
phryneas, line 170: This is now throttled. Before, events leading up to `documentDidChange` were debounced in `server.ts`, but that lead to file updates being completely lost if more than one document changed.

src/language-server/project/internal.ts
```diff
@@ -0,0 +1,346 @@
...
+  ".svelte": "svelte",
+  ".py": "python",
+  ".rb": "ruby",
+  ".dart": "dart",
+  ".re": "reason",
+  ".ex": "elixir",
+  ".exs": "elixir",
+};
+
+export interface GraphQLInternalProjectConfig extends GraphQLProjectConfig {
```
phryneas, line 51: Most of this just moved over from `base.ts`

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -0,0 +1,233 @@
...
+    }
+  }
+  return newParts;
+}
+
+function getUri(part: FilePart) {
+  return part.source.name + "/" + part.fractionalIndex + ".graphql";
+}
+
+export class DocumentSynchronization {
```
phryneas, line 71: This class slows down incoming document updates, and splits them up into multiple "partial documents" if we e.g. have a `.js` file with multiple `gql`-tagged blocks.

Calling `synchronizedWithDocument` will ensure that changes to a specific document are flushed if an action like e.g. hover is forwarded to the upstream LSP.

`insideVirtualDocument` calls `synchronizedWithDocument` on a parent document and then executes the callback with a "virtual" file inside that document.

src/language-server/project/rover/project.ts
```diff
@@ -0,0 +1,201 @@
...
+export function isRoverConfig(config: ApolloConfig): config is RoverConfig {
+  return config instanceof RoverConfig;
+}
+
+export interface RoverProjectConfig extends GraphQLProjectConfig {
+  config: RoverConfig;
+  capabilities: ClientCapabilities;
+}
+
+export class RoverProject extends GraphQLProject {
```
phryneas, line 35: The class for new LSP projects.

src/language-server/server.ts
```diff
@@ -142,21 +146,30 @@ function isFile(uri: string) {
   return URI.parse(uri).scheme === "file";
 }
 
-documents.onDidChangeContent(
-  debounceHandler((params) => {
```
phryneas, line 146: This debounce was the source of a potential bug.

src/language-server/utilities/__tests__/source.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  });
+
+  test("should return the correct position on the first line of a document", () => {
+    expect(
+      positionFromPositionInContainingDocument(sources[0], {
+        line: 4,
+        character: 4,
+      }),
+    ).toEqual({ line: 0, character: 0 });
+  });
```
phryneas, line 45: Test for fixed behaviour

sampleWorkspace/rover/src/test.js
```diff
@@ -0,0 +1,30 @@
+import gql from "graphql-tag";
+
+sdfsdfs;
```
jerelmiller, line 3: Was this intentionally added? If so, could you add a comment explaining why its here?

sampleWorkspace/rover/src/test.js
```diff
@@ -0,0 +1,30 @@
...
+  """
+  Test
+  """
+  type User {
+    id: ID!
+    name: String!
+  }
+`;
+
+console.log("foobar!");
```
jerelmiller, line 21: Is this needed?

src/language-server/config/config.ts
```diff
@@ -79,7 +81,29 @@ const clientConfig = z.object({
 export type ClientConfigFormat = z.infer<typeof clientConfig>;
 
 const roverConfig = z.object({
-  bin: z.string().optional(),
+  bin: z
+    .preprocess(
+      (val) => val || which.sync("rover", { nothrow: true }) || undefined,
+      z.string({
+        message:
+          "Rover binary not found. Please either install it system-wide in PATH, or provide the `bin` option. Please also ensure that the binary is executable.",
```
jerelmiller:
```suggestion
          "Rover binary not found. Please install it system-wide in PATH or provide the `bin` option and ensure the binary is executable.",
```

Feel free to tweak this further, but I'd recommend combining both sentences to remove the double "please" call-to-action here.

src/language-server/config/config.ts
```diff
@@ -79,7 +81,29 @@ const clientConfig = z.object({
...
+          // is executable?
+          accessSync(bin, fsConstants.X_OK);
+          // is a file and not a directory?
+          return statSync(bin).isFile();
+        } catch {
+          return false;
+        }
+      },
+      {
+        message: "Rover binary is not executable",
```
jerelmiller:
```suggestion
        message: "Rover binary is installed but not not executable. Please ensure the rover binary is executable for...",
```

I'm not quite sure how to finish this sentence, but would it sense to provide a call to action here? At the very least, I'd consider mentioning that you were able to find the binary, but just couldn't execute it.

src/language-server/utilities/__tests__/source.test.ts
```diff
@@ -0,0 +1,115 @@
+import { extractGraphQLSources } from "../../document";
+import { TextDocument } from "vscode-languageserver-textdocument";
+import {
+  findContainedSourceAndPosition,
+  positionFromPositionInContainingDocument,
+} from "../source";
+import { handleFilePartUpdates } from "../../project/rover/DocumentSynchronization";
+const testText = `import gql from "graphql-tag";
```
jerelmiller:
```suggestion
import { handleFilePartUpdates } from "../../project/rover/DocumentSynchronization";

const testText = `import gql from "graphql-tag";
```

Minor thing, but my eyes initially thought the test text was part of the set of import statements above 😆

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -0,0 +1,233 @@
...
+    if (!found) {
+      return;
+    }
+    const match = findContainedSourceAndPosition(
+      found.parts,
+      positionParams.position,
+    );
+
+    if (!match) return;
+    console.log(positionParams.position, match.position);
```
jerelmiller:
```suggestion
```

src/language-server/project/rover/project.ts
```diff
@@ -0,0 +1,190 @@
...
+    super(options);
+    this.config = options.config;
+    this.capabilities = options.capabilities;
+  }
+
+  initialize() {
+    return [this.connection.then(() => {})];
+  }
+
+  get connection(): Promise<ProtocolConnection> {
```
jerelmiller, line 56: I'm curious, is this used by the superclass at all? If not, I tend to prefer `async getConnection()` here and make this an actual method, mostly because I think it helps a touch when reading it in lines like this:

```ts
const connection = await this.getConnection();
```

Without making it a function, it looks both redundant and a bit confusing why we await the property:

```ts
const connection = await this.connection;
```

I'm happy to leave this as-is too, just figured this might make it a touch clearer whats happening here.

src/language-server/project/rover/project.ts
jerelmiller, line 1: I'll add a general comment for this whole file.

Is every single `console.log` in here necessary? Some of them look useful and makes sense to keep, but some of these look more like they were leftover from debugging.

I trust you to leave the ones that make sense, but I'd just advise looking through here to double check which logs should be kept/removed.

jerelmiller (approved):
Most of my comments were minor things. I won't pretend to understand this 100% in depth, but the tests were easy to follow and made it easier to understand what was happening. Thanks for structuring those the way you did!

sampleWorkspace/rover/src/test.js
phryneas: Padding the file with *something* that's not only `gql` tags - of course I could also copy-paste over random code, but for now this seemed to do ^^

sampleWorkspace/rover/src/test.js
phryneas: Same as above - this file (as everything in sampleWorkspace) is only meant to open as an "example workspace" so we can see where highlighting works.

If the whole `.js` file would consist out of `gql` tags, it would be pointless - there has to be some filler. But the file will never be executed, it purely exists to look at it in the editor.

src/language-server/config/config.ts
phryneas: The "executable" part applies to both "or paths" of the sentence before, so it's `(in PATH or bin option) AND executable. It seemed less ambiguous to me to keep them separated - but I can remove the second `please`.
```suggestion
          "Rover binary not found. Please either install it system-wide in PATH, or provide the `bin` option. Also ensure that the binary is executable.",
```

src/language-server/config/config.ts
phryneas:
```suggestion
        message: "Rover binary is not marked as an executable. If you are using OS X or Linux, ensure to set the executable bit.",
```

src/language-server/project/rover/project.ts
phryneas: A lot of them go away in the future, while some end up in a `DEBUG` mode, but for now it's good I see everything that's happening there - and it being behind a feature flag, nobody else will ;)

src/language-server/project/rover/project.ts
phryneas: I like the suggestion, but I'll do it in the next PR :)

PR #176, opened 2024-08-19
upstream LSP: basic diagnostic functionality

phryneas:
This 
* fixes a position mapping bug
* adds diagnostics support
* hardens logic around project recreation, recovery after upstream LSP crashes and similar


src/language-server/project/base.ts
```diff
@@ -131,6 +133,14 @@ export abstract class GraphQLProject {
   abstract onCodeLens?: ConnectionHandler["onCodeLens"];
   abstract onCodeAction?: ConnectionHandler["onCodeAction"];
 
+  abstract onUnhandledRequest?: StarRequestHandler;
+  abstract onUnhandledNotification?: (
+    connection: Connection,
```
phryneas, line 138: Deviating from the `StarNotificationHandler` signature and prepending `connection`, so the handler could actually do something in response to the notification if necessary.
(Right now I'm just using these handlers for logging though, this is more of a future-proofing)

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -5,23 +5,30 @@ import {
...
   source: Source;
+  diagnostics: Diagnostic[];
 }
 
 export function handleFilePartUpdates(
-  parsed: Source[],
-  previousParts: FilePart[],
-): FilePart[] {
-  const newParts = [];
+  parsed: ReadonlyArray<Source>,
```
phryneas, line 28: Changed all arrays to `ReadonlyArray` here because I want to prevent any accidental mutation. Should have been like this from the start.

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -37,7 +44,7 @@ export function handleFilePartUpdates(
         newOffset === oldPart.source.locationOffset.line + offsetCorrection)
     ) {
       // replacement of chunk
-      newParts.push({ source, fractionalIndex: oldPart.fractionalIndex });
+      newParts.push({ ...oldPart, source });
```
phryneas, line 47: Keeping the old file part around and just updating the new `source`. This is important as file parts now have an additional `diagnostics` property that should persist user changes until the upstream LSP gets to reporting new diagnostics.

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -67,14 +74,20 @@ export function handleFilePartUpdates(
 function getUri(part: FilePart) {
   return part.source.name + "/" + part.fractionalIndex + ".graphql";
 }
+function splitUri(fullUri: DocumentUri) {
```
phryneas, line 77: "reverse `getUri`"

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -109,10 +123,12 @@ export class DocumentSynchronization {
     }
   };
 
-  private async sendDocumentChanges(document: TextDocument) {
+  private async sendDocumentChanges(
+    document: TextDocument,
+    previousParts = this.knownFiles.get(document.uri)?.parts || [],
```
phryneas, line 128: `previousParts` can now be passed as an argument, as in the case of a restart of the upstream LSP we want to re-send the full document as a series of `didOpen` notifications using this function.

(See the newly added `resendAllDocuments`)

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -229,4 +251,48 @@ export class DocumentSynchronization {
       position: match.position,
     });
   }
+
+  handlePartDiagnostics(params: PublishDiagnosticsParams) {
```
phryneas, line 255: Diagnostics from the upstream LSP are for "virtual files" like `${realFileName}/${index}` - and we have to map these to the correct document part, replace the diagnostics of that part with the reported diagnostics, and then report the diagnostics for the full file (accumulation of virtual files) back to VSCode.

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -229,4 +251,48 @@ export class DocumentSynchronization {
...
+    };
+
+    this.sendDiagnostics(fullDocumentParams);
+  }
+
+  get openDocuments() {
+    return [...this.knownFiles.values()].map((f) => f.full);
+  }
+
+  clearAllDiagnostics() {
```
phryneas, line 290: If the project is discarded (or recreated), we need to tell VSCode to reset all diagnostics reported by it - after all, the `vscode-graphql` LSP is just a facade in front of many Client or Rover projects, and VSCode doesn't know what belongs to which of those projects.

src/language-server/project/rover/project.ts
```diff
@@ -50,56 +67,93 @@ export class RoverProject extends GraphQLProject {
...
+    if (this.disposed) {
+      throw new ConnectionError(
+        ConnectionErrors.Closed,
+        "Connection is closed.",
+      );
+    }
     const child = cp.spawn(this.config.rover.bin, ["lsp"], {
-      env: { RUST_BACKTRACE: "1" },
+      env: DEBUG ? { RUST_BACKTRACE: "1" } : {},
+      stdio: ["pipe", "pipe", DEBUG ? "inherit" : "ignore"],
```
phryneas, line 148: `"inherit"` here would just re-print STDERR to the parent process STDERR - it replaces 

```
    child.stderr.on("data", (data) => {
      console.info("stderr", data.toString());
    });
```

src/language-server/project/rover/project.ts
```diff
@@ -108,30 +162,40 @@ export class RoverProject extends GraphQLProject {
       console.error({ err });
     });
 
+    connection.onNotification(
+      PublishDiagnosticsNotification.type,
+      this.documents.handlePartDiagnostics.bind(this.documents),
```
phryneas, line 167: Trying hard to get into the habit of using `.bind` instead of arrow functions where possible to prevent the risk of closure-related memory leaks 🙈

src/language-server/project/rover/project.ts
```diff
@@ -108,30 +162,40 @@ export class RoverProject extends GraphQLProject {
...
-        (error) => {
-          console.error("Connection failed to initialize", error);
-          throw error;
-        },
       );
+      DEBUG && console.log("Connection initialized", status);
+
+      await this.connectionStorage.run(connection, () =>
+        this.documents.resendAllDocuments(),
+      );
```
phryneas: `resendAllDocuments` calls `project.sendNotification`, which calls `this.getConnection()`, which then again would latch onto the `this._connection` promise, which will only resolve after the `initializeConnection` function we currently are in finishes.

Using `this.connectionStorage`, an `AsyncLocalStorage` to overwrite the result of `this.getConnection()` for this one async code path temporarily.

That way we execute with higher priority than any earlier-scheduled `project.sendNotification` call that might depend on the result state of `resendAllDocuments`.

src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -67,14 +74,20 @@ export function handleFilePartUpdates(
 function getUri(part: FilePart) {
   return part.source.name + "/" + part.fractionalIndex + ".graphql";
 }
+function splitUri(fullUri: DocumentUri) {
+  const result = /^(.*)\/(\w+)\.graphql/.exec(fullUri);
```
jerelmiller, line 78: Does this regex need to handle (what I think are) filenames with intermediate dots in them? For example: `schema.local.graphql`. I see this is called from `handlePartDiagnostics`, but I'm unsure what the original `uri` is passed to that function.

src/language-server/project/rover/project.ts
```diff
@@ -17,11 +16,18 @@ import {
...
+  ConnectionErrors,
 } from "vscode-languageserver/node";
 import cp from "node:child_process";
 import { GraphQLProjectConfig } from "../base";
 import { ApolloConfig, RoverConfig } from "../../config";
 import { DocumentSynchronization } from "./DocumentSynchronization";
+import { AsyncLocalStorage } from "node:async_hooks";
+import internal from "node:stream";
+
+const DEBUG = true;
```
jerelmiller, line 30: Will the user see the messages logged when this is true? Does this need to be toggleable somehow? I'm curious the reason why this makes sense to always be enabled vs just logging the messages directly without this check.

src/language-server/project/rover/project.ts
jerelmiller:
```suggestion
      await this.connectionStorage.run(connection, this.documents.resendAllDocuments.bind(this.documents);
```

Since you're wanting to avoid closure-related memory leaks 🙂

jerelmiller (approved):
Changes look good to me. Mostly had a couple questions.

src/language-server/project/rover/DocumentSynchronization.ts
phryneas: That last part would always be the `fractionalIndex`, so something like `"a0"`, `"a1"` or `"a0y"`, so this should be enough. That said, #180 will move this around again anyways :)

src/language-server/project/rover/project.ts
phryneas: Right now it's all behind a featureflag - so before we ship it without the feature flag, I'll probably replace that static `true` with another feature flag or something user-configurable. I just want this here so I can already put things "behind it".

PR #179, opened 2024-08-22
Adjust TS/JS regexp for Document detection, also adjust TextMate

phryneas:
Before:
image
After:
image
Before:
image
After:
image
Before:
image
After:
image
Before:
image
After:
image


src/language-server/__tests__/fixtures/documents/commentWithTemplate.ts
```diff
@@ -0,0 +1,41 @@
...
+/*GraphQL*/ `query Q4 {test}`;
+/** GraphQL */ `query Q5 {test}`;
+
+const _3 = /* GraphQL */ `
+  query Q6 {
+    test
+  }
+`;
+
+// syntax highlighting cannot work in all examples after this - textmate grammars don't work over multiple lines like this
```
phryneas, line 28: While we cannot syntax highlight anything after that line, the other features will still work. We *could* disable them with stricter regular expressions, but I'm not sure if we should.

src/language-server/__tests__/document.test.ts
```diff
@@ -119,6 +174,101 @@ describe("extractGraphQLDocuments", () => {
...
+        .toMatchInlineSnapshot(`
+Array [
+  "Q1",
+  "Q2",
+  "Q3",
+  "Q4",
+  "Q6",
+  "Q8",
+  "Q9",
+  "Q10",
```
phryneas, line 227: We just test that all queries in the fixture are found by looking at the name, and the `documentName` function also checks that there were no syntax errors in the document.

src/language-server/__tests__/fixtures/documents/commentWithTemplate.ts.snap
```diff
@@ -0,0 +1,185 @@
...
+#^^^^^ source.ts meta.block.ts meta.var.expr.ts storage.type.ts
+#     ^ source.ts meta.block.ts meta.var.expr.ts
+#      ^^ source.ts meta.block.ts meta.var.expr.ts meta.var-single-variable.expr.ts meta.definition.variable.ts variable.other.constant.ts
+#        ^ source.ts meta.block.ts meta.var.expr.ts meta.var-single-variable.expr.ts
+#         ^ source.ts meta.block.ts meta.var.expr.ts keyword.operator.assignment.ts
+#          ^ source.ts meta.block.ts meta.var.expr.ts
+#           ^^^ source.ts meta.block.ts meta.var.expr.ts
+#              ^^^^^^^ source.ts meta.block.ts meta.var.expr.ts markup.italic
+#                     ^^^^ source.ts meta.block.ts meta.var.expr.ts
+#                         ^^^^^ source.ts meta.block.ts meta.var.expr.ts meta.embedded.block.graphql keyword.operation.graphql
```
phryneas, line 36: This shows that everything above the `^^^` here was detected as both `meta.embedded.block.graphql` and `keyword.operation.graphql`

src/language-server/__tests__/fixtures/documents/commentWithTemplate.ts.snap
```diff
@@ -0,0 +1,185 @@
...
+#^ source.ts meta.block.ts punctuation.definition.block.ts
+>const _1 = /* GraphQL */`query Q1 {test}`;
+#^^^^^ source.ts meta.block.ts meta.var.expr.ts storage.type.ts
+#     ^ source.ts meta.block.ts meta.var.expr.ts
+#      ^^ source.ts meta.block.ts meta.var.expr.ts meta.var-single-variable.expr.ts meta.definition.variable.ts variable.other.constant.ts
+#        ^ source.ts meta.block.ts meta.var.expr.ts meta.var-single-variable.expr.ts
+#         ^ source.ts meta.block.ts meta.var.expr.ts keyword.operator.assignment.ts
+#          ^ source.ts meta.block.ts meta.var.expr.ts
+#           ^^^ source.ts meta.block.ts meta.var.expr.ts
+#              ^^^^^^^ source.ts meta.block.ts meta.var.expr.ts markup.italic
```
phryneas, line 34: I've chosen to make the word `GraphQL` in the `/* GraphQL */` comment here italic to highlight that it does something.

syntaxes/graphql.js.json
```diff
@@ -3,43 +3,40 @@
...
+    {
+      "contentName": "meta.embedded.block.graphql",
+      "begin": "`(\\s*#[ ]*(gql|graphql|GraphQL))",
       "beginCaptures": {
         "1": {
-          "name": "entity.name.function.tagged-template.js"
+          "name": "meta.embedded.block.graphql comment.line.graphql.js"
         },
         "2": {
-          "name": "punctuation.definition.string.template.begin.js"
```
phryneas, line 14: I did remove these `entity.name.function.tagged-template.js` and `punctuation.definition.string.template.begin.js` since we don't want *our* grammar to attack tokens and highlight those, but we want the parent language to add them.
Instead, for only this pattern, we add `meta.embedded.block.graphql comment.line.graphql.js"`, as the `source.graphql` matcher will only start working between start and end pattern, but we also want the GraphQL in the start pattern to be properly highlighted.

The other patterns don't add capture names, apart from `markup.italic` on the word `gql/graphql/GraphQL` inside of comments.

syntaxes/graphql.js.json
```diff
@@ -3,43 +3,40 @@
   "injectionSelector": "L:source -string -comment",
   "patterns": [
```
phryneas, line 4: This is split into four patterns now instead of one pattern that kinda tries to capture everything with a monster RegExp.

syntaxes/graphql.js.json
```diff
@@ -3,43 +3,40 @@
...
     {
-      "name": "taggedTemplates",
```
phryneas, line 6: That name was applied to all the captures, but didn't seem to do anything. It's optional, so I removed it.

syntaxes/graphql.js.json
```diff
@@ -3,43 +3,40 @@
...
       "contentName": "meta.embedded.block.graphql",
-      "begin": "(Relay\\.QL|gql|graphql(\\.experimental)?)\\(?(`)",
+      "begin": "(?:Relay\\.QL|gql|graphql(\\.experimental)?)\\s*(?:<.*?>\\s*)?`",
```
phryneas, line 7: We now also allow for TypeScript generics here - `(?:<.*?>\\s*)?`.
That might come in handy if we add those in a future version of `graphql-tag`

src/language-server/document.ts
```diff
@@ -98,20 +114,30 @@ function extractGraphQLSourcesFromJSTemplateLiterals(
...
   );
 
   let result;
   while ((result = regExp.exec(text)) !== null) {
-    const contents = replacePlaceholdersWithWhiteSpace(result[1]);
-    const position = document.positionAt(result.index + (tagName.length + 1));
+    // we have multiple alternative capture groups in the regexp, and only one of them will have a result
+    // so we need the index for that
+    const groupIndex = result.findIndex(
+      (part, index) => index != 0 && part != null,
```
jerelmiller:
```suggestion
      (part, index) => index !== 0 && part != null,
```

jerelmiller (approved):
Spotted only one minor thing, but otherwise looks great! Cool to see how some of the syntax highlighting works behind the scenes.

PR #183, opened 2024-08-26
[upstream LSP] support for semantic tokens

phryneas:


src/language-server/project/rover/DocumentSynchronization.ts
```diff
@@ -305,4 +315,71 @@ export class DocumentSynchronization {
...
+          ),
+        };
+      }),
+    );
+    let line = 0,
+      char = 0,
+      lastLine = 0,
+      lastChar = 0;
+    const combinedTokens = [];
+    for (const { part, tokens } of allParts) {
```
phryneas, line 357: `tokens` here looks like this:
```
{
    data: [
      10, 20, 1, 2, 0, 0, 1,  1, 0,
       0,  0, 1, 1, 2, 0, 2, 31, 1,
       2,  0, 0, 1, 1, 0, 0,  0, 1,
       1,  2, 0
    ]
}
```

src/language-server/project/base.ts
```diff
@@ -120,29 +121,32 @@ export abstract class GraphQLProject {
   abstract includesFile(uri: DocumentUri): boolean;
 
   abstract onDidChangeWatchedFiles: ConnectionHandler["onDidChangeWatchedFiles"];
-  abstract onDidOpen?: (event: TextDocumentChangeEvent<TextDocument>) => void;
-  abstract onDidClose?: (event: TextDocumentChangeEvent<TextDocument>) => void;
```
phryneas, line 127: Removing the `abstract` here for the optional methods means we can remove the weird `undefined` definitions where they are not implemented.

src/language-server/project/rover/project.ts
```diff
@@ -250,8 +257,17 @@ export class RoverProject extends GraphQLProject {
       this.sendRequest(HoverRequest.type, virtualParams, token),
     );
 
-  onUnhandledRequest: GraphQLProject["onUnhandledRequest"] = (type, params) => {
-    DEBUG && console.info("unhandled request from VSCode", { type, params });
+  onUnhandledRequest: GraphQLProject["onUnhandledRequest"] = async (
```
phryneas, line 273: This is in `onUnhandledRequest` since adding listeners in `server.ts` for this feature would be clumsy anyway (it's not a property like `connection.onSemanticTokenRequest`, but `connection.languages.semanticTokens.on` - yes, literally `on` 🤦 

So it seems more readable to do it just here.

jerelmiller (approved):
💯

PR #184, opened 2024-08-27
Derive extensions for supported languages and monitored files from other installed extensions.

phryneas:


src/language-server/fileSet.ts
```diff
@@ -14,12 +14,10 @@ export class FileSet {
     rootURI,
     includes,
     excludes,
-    configURI,
```
phryneas, line 17: This was never used.

src/language-server/fileSet.ts
```diff
@@ -30,13 +28,6 @@ export class FileSet {
     this.excludes = excludes;
   }
 
-  pushIncludes(files: string[]) {
-    this.includes.push(...files);
-  }
-  pushExcludes(files: string[]) {
-    this.excludes.push(...files);
-  }
```
phryneas, line 38: These are not used anymore since `fileSet` has been split into `GraphQLProject.configFileSet` and `InternalGraphQLProject.fileSet`

src/language-server/project/client.ts
```diff
@@ -201,20 +201,7 @@ export class GraphQLClientProject extends GraphQLInternalProject {
...
-     * and there's no way to know for sure that there _should_ be files.
-     */
-    const filterConfigAndEnvFiles = (path: string) =>
-      !(
-        path.includes("apollo.config") ||
-        path.includes(".env") ||
-        (config.configURI && path === config.configURI.fsPath)
-      );
-
-    if (this.allIncludedFiles().filter(filterConfigAndEnvFiles).length === 0) {
```
phryneas, line 217: Since `this.fileSet` doesn't contain config files anymore, this is not necessary anymore.

src/language-server/project/rover/project.ts
```diff
@@ -38,6 +40,10 @@ export interface RoverProjectConfig extends GraphQLProjectConfig {
   capabilities: ClientCapabilities;
 }
 
+const supportedLanguageIds = (
+  process.env.APOLLO_ROVER_LANUGAGE_IDS || "graphql"
+).split(",");
```
phryneas: This resricts the rover LSP to only `.graphql`/`.gql`/etc. files and disables it for `.js` files and similar for now until we have a good user story there.

src/language-server/server.ts
```diff
@@ -84,43 +89,49 @@ workspace.onConfigFilesFound(async (params) => {
...
-    // We wait until all projects are added, because after `initialize` returns we can get additional requests
-    // like `textDocument/codeLens`, and that way these can await `GraphQLProject#whenReady` to make sure
-    // we provide them eventually.
-    await Promise.all(
-      workspaceFolders.map((folder) => workspace.addProjectsInFolder(folder)),
+connection.onInitialize(
+  async ({ capabilities, workspaceFolders, initializationOptions }) => {
+    const { languageIdExtensionMap } =
+      initializationOptions as InitializationOptions;
+    setLanguageIdExtensionMap(languageIdExtensionMap);
```
phryneas, line 96: This is really all that changed here, the rest is whitespace.

src/language-server/server.ts
```diff
@@ -246,28 +264,31 @@ connection.onWorkspaceSymbol(async (params, token) => {
 
 connection.onCompletion(
   debounceHandler(
-    (params, token, workDoneProgress, resultProgress) =>
+    ((params, token, workDoneProgress, resultProgress) =>
```
phryneas, line 267: Before this change, all of these were implicit any 🤦

src/tools/utilities/getLanguageInformation.ts
```diff
@@ -0,0 +1,40 @@
+import * as vscode from "vscode";
+import {
+  LanguageIdExtensionMap,
+  minimumKnownExtensions,
+} from "./languageInformation";
+
+/**
+ * @returns An object with language identifiers as keys and file extensions as values.
```
phryneas, line 8:
```suggestion
 * @returns An object with language identifiers as keys and file extensions as values.
 * see https://github.com/microsoft/vscode/issues/109919
```

src/language-server/project/rover/project.ts
```diff
@@ -38,6 +40,10 @@ export interface RoverProjectConfig extends GraphQLProjectConfig {
   capabilities: ClientCapabilities;
 }
 
+const supportedLanguageIds = (
+  process.env.APOLLO_ROVER_LANUGAGE_IDS || "graphql"
```
jerelmiller:
```suggestion
  process.env.APOLLO_ROVER_LANGUAGE_IDS || "graphql"
```

Typo. Not sure if this affects other places, but you may want to do a quick grep of the codebase to check 🙂

src/language-server/project/rover/project.ts
```diff
@@ -203,8 +209,14 @@ export class RoverProject extends GraphQLProject {
     return { type: "Rover", loaded: true };
   }
 
-  includesFile(uri: DocumentUri) {
-    return uri.startsWith(this.rootURI.toString());
+  includesFile(
+    uri: DocumentUri,
+    languageId = getLanguageIdForExtension(extname(uri) as `.${string}`),
```
jerelmiller: I've seen the `.${string}` type repeated a couple times in this PR. Would it be worth extracting an `Extension`/`FileExtension` type somewhere?

i.e.
```ts
type FileExtension = `.${string}`
```

jerelmiller (approved):
Spotted a typo, but otherwise looks great!

src/language-server/project/rover/project.ts
phryneas: Good call, thank you!

src/language-server/project/rover/project.ts
phryneas: Good call, done!

PR #196, opened 2024-09-10
Default to `supergraph.yaml` instead of `.yml`

dylan-apollo:
All of our examples/docs use the full extension, so this tripped me up today. We could optionally, also, check both extensions.


phryneas (approved):

phryneas:
We can check for both in the long run, but let's get started with this :)

PR #197, opened 2024-09-11
Adjustments to VSCode <-> DevTools communication

phryneas:
This is the sister PR for https://github.com/apollographql/apollo-client-devtools/pull/1513

CI fails because no devtools package has been published yet, but please review it anyways :)

### DIFF
--- a/src/devtools/DevToolsViewProvider.ts
+++ b/src/devtools/DevToolsViewProvider.ts
@@ -1,5 +1,34 @@
 import * as vscode from "vscode";
-import { devtoolsEvents } from "./server";
+import { devtoolsEvents, sendToDevTools, serverState } from "./server";
+
+type ActorMessage = { type: "actor"; message: unknown };
+
+export function isActorMessage(message: any): message is ActorMessage {
+  return message && message.type === "actor";
+}
+
+type DevToolsOpenCommandMessage = {
+  type: "vscode:executeCommand";
+  command: string;
+  arguments?: unknown[];
+};
+
+export function isDevToolsExecuteCommandMessage(
+  message: any,
+): message is DevToolsOpenCommandMessage {
+  return message && message.type === "vscode:executeCommand";
+}
+
+type DevToolsOpenExternalMessage = {
+  type: "vscode:openExternal";
+  uri: string;
+};
+
+export function isDevToolsOpenExternalMessage(
+  message: any,
+): message is DevToolsOpenExternalMessage {
+  return message && message.type === "vscode:openExternal";
+}
 
 export class DevToolsViewProvider implements vscode.WebviewViewProvider {
   public static readonly viewType = "vscode-apollo-client-devtools";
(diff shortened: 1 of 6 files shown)


src/devtools/DevToolsViewProvider.ts
```diff
@@ -18,16 +48,40 @@ export class DevToolsViewProvider implements vscode.WebviewViewProvider {
...
+                  serverState?.port ||
+                  vscode.workspace
+                    .getConfiguration("apollographql")
+                    .get("devTools.serverPort", 0),
+                listening: !!serverState?.port,
+              },
+            });
+          }
+        }
+      },
```
phryneas, line 73: Where previously the panel kickstarted itself, it now notifies the extension, and the extension dispatches the `initializePanel` message with some additional context.

src/devtools/server.ts
```diff
@@ -15,8 +15,53 @@ devtoolsEvents.addListener("fromDevTools", (msg) => {
...
+    type: "actor",
+    message,
+  });
+}
+
+export let serverState:
+  | { port: false | number; disposable: Disposable }
+  | undefined = undefined;
+
+export function startServer(port: number) {
```
phryneas, line 33: A bunch of checks moved over from `extension.ts` here and the whole "disposing" and "restarting" should be a lot more solid now.

src/devtools/server.ts
```diff
@@ -15,8 +15,53 @@ devtoolsEvents.addListener("fromDevTools", (msg) => {
...
+    id: `vscode-${id++}`,
+    source: "apollo-client-devtools",
+    type: "actor",
+    message,
+  });
+}
+
+export let serverState:
+  | { port: false | number; disposable: Disposable }
+  | undefined = undefined;
```
phryneas, line 31: `false` if the server is in the middle of starting, `number` if the server is running. The whole `serverState` is `undefined` after the server has stopped or if it has never been started before.

src/devtools/DevToolsViewProvider.ts
```diff
@@ -20,16 +49,40 @@ export class DevToolsViewProvider implements vscode.WebviewViewProvider {
...
-      devtoolsEvents.emit("fromDevTools", data);
-    });
+    const panelDisposables: vscode.Disposable[] = [];
+    panel.webview.onDidReceiveMessage(
+      (data) => {
+        if (data.source === "apollo-client-devtools") {
+          devtoolsEvents.emit("fromDevTools", data);
+        }
+        if (data.source === "vscode-panel") {
+          if (data.type === "mounted") {
```
jerelmiller, line 59:
```suggestion
        if (data.source === "vscode-panel" && data.type === "mounted") {
```

You should be able to combine these conditionals.

src/devtools/server.ts
jerelmiller: Out of curiosity, why use `false` instead of `null` or `undefined` to represent the initial state?

src/devtools/server.ts
```diff
@@ -15,8 +15,53 @@ devtoolsEvents.addListener("fromDevTools", (msg) => {
...
+
+export function startServer(port: number) {
+  const state = {
```
jerelmiller, line 34: I'm curious, whats the advantage of using a separate local `state` variable here instead of just assigning this object to `serverState` and modifying that in `wss.on('listening')` for example? Is it to avoid `?` or `!` since `serverState` is of type `| undefined`?

src/devtools/server.ts
```diff
@@ -15,8 +15,53 @@ devtoolsEvents.addListener("fromDevTools", (msg) => {
...
+  };
+  if (serverState) {
+    if (serverState.port === port) {
+      // nothing to do
+      return;
+    } else {
+      // changing port, stop the old server
+      serverState.disposable.dispose();
+    }
+  }
```
jerelmiller:
```suggestion
  if (serverState && serverState.port !== port) {
    // changing port, stop the old server
    serverState.disposable.dispose();
  }
```

Should be able to simplify this by combining the conditionals and getting rid of the branch that doesn't do anything.

jerelmiller (approved):
Had a few suggestions to tighten up the code a bit, but otherwise looks good

src/devtools/server.ts
jerelmiller: I just noticed the `return` keyword 🤦. Ignore the original suggestion.

That said, I think just adding an `if` above this makes sense:

```ts
if (serverState?.port === port) {
  return;
}
```

Which would mean that any code after this would mean that `serverState.port` is different or uninitialized. I think then you can just remove the `if`. The whole thing would look like:

```ts
if (serverState?.port === port) {
  return;
}

// changing port, stop the old server
serverState?.disposable.dispose();
// ...
```

src/devtools/DevToolsViewProvider.ts
phryneas: I assume that inside of `data.source` we'll see more reactions to other `data.type` in the future, which is why I separated those from each other.

src/devtools/server.ts
phryneas: It felt more deliberate than `undefined`. `null` would have been fine too, I guess.

src/devtools/server.ts
phryneas: See a bit further down: `if (serverState === state) {` - it makes it possible to detect if `serverState` has been reassigned.

src/devtools/server.ts
phryneas: c902746991f5d7fa339342b41226f05cb71ef4fb ✅

phryneas: and d4a65ee8cc6a0f82a3a83c43019c16ee8351bfab ... that took a moment to click for me

PR #198, opened 2024-09-11
"graphql-object-field": highlighting "string" -> "variable"

phryneas:
Prioritizes "constant.object.key.graphql" over "string.unquoted.graphql", changes "constant" to "variable".

This makes nested objects in graphql arguments look more in-line.

Snippet for testing:

```graphql
type Query {
  users: [User] @auth(
    role: User,
    permissions: {
      write: true
#     ^^^^^ funny how GitHub also highlights this as a string here
    }
  )
}

enum Role {
  User
  Admin
}

type User {
  id: ID!
  name: String!
}

input Permissions {
  read: Boolean
  write: Boolean
}

directive @auth(role: Role, permissions: Permissions) on FIELD_DEFINITION
```

Without these changes:

image

With these changes:

image

Note that after these changes `write` is highlighted the same as `role` or `permissions`.

### DIFF
--- a/.changeset/chatty-toes-hammer.md
+++ b/.changeset/chatty-toes-hammer.md
@@ -1,0 +1,5 @@
+---
+"vscode-apollo": patch
+---
+
+Change syntax highlighting of graphql object fields from "string" to "variable".
--- a/syntaxes/graphql.json
+++ b/syntaxes/graphql.json
@@ -601,8 +601,8 @@
     "graphql-object-field": {
       "match": "\\s*(([_A-Za-z][_0-9A-Za-z]*))\\s*(:)",
       "captures": {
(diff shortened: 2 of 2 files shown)


dylan-apollo (approved):
Based on the naming, I bet the reason this was done this way was for inline input objects in _queries_:

```graphql
query Blah {
  something(input: {var: "thing"})
}
```

The distinction between the argument name `input` and the values within the object could be useful.

As long as we're okay with changing _both_ to be consistent, which is how the JetBrains extension looks, so _I'm_ used to, then this looks good to me!

phryneas:
Honestly, with that explanation, this looks even better to me. It's very weird that in your GH comment, `var` is string-colored.

This is also more in line with how object properties in JavaScript are highlighted:
image

PR #202, opened 2024-09-13
Prevent configuration file "loading loop" when transpiling TS.

phryneas:
This was also present in the original `cosmiconfig` loader logic, which would compile a `apollo.config.ts` to a `apollo.config.mjs` and then import from that file.

Problem with this is that that file creation (and deletion) would trigger the file event watcher, which then would trigger another config parse, leading to an infinite loop. 
This didn't lead to particularly high system load, since it happened only every 2 seconds or so, with all the async layers involved, but it was definitely problematic.

We now create that file in a temporary directory outside the workspace, and we use `apollo.config.ts.mjs`, which is not a valid config file name and wouldn't lead to a loop even if picked up (e.g. if someone opens the system temp folder as a workspace).


src/language-server/workspace.ts
```diff
@@ -215,9 +215,6 @@ export class GraphQLWorkspace {
     }
     // If project exists, update the config
     if (project && config) {
-      if (equal(project.config.rawConfig, config.rawConfig)) {
-        return;
-      }
```
phryneas, line 220: I added this logic when I saw that there were too many file events happening, but didn't know why and assumed the reason was something outside the scope of this Extension.

Removing it here again, since sometimes we actually want a restart even if the config didn't really change.

phryneas:
I've come to the conclusion that creating this file in another folder is not a good idea, since it would stop any imports from working... doing some changes.

jerelmiller (approved):

PR #204, opened 2024-09-13
Cache busting config reloading by using virtual files

phryneas:
We needed a way to get around node's `import` cache, so this seems to be the best way to do that. Setting a module-path query parameter doesn't seem to have any effect.

On a positive sidenote, we also don't have to write any files to disk for TS transpilation anymore.


src/language-server/server.ts
```diff
@@ -192,9 +192,6 @@ documents.onDidClose(
 connection.onDidChangeWatchedFiles((params) => {
   const handledByProject = new Map<GraphQLProject, FileEvent[]>();
   for (const { uri, type } of params.changes) {
-    if (uri.match(/vscode-extension-ignore.[mc]?js$/)) {
-      continue;
-    }
```
phryneas, line 197: This is no longer required since we don't create files on disk anymore

src/language-server/config/loadTsConfig.ts
```diff
@@ -30,36 +39,40 @@ export const loadTs: Loader = async function loadTs(filepath, content) {
...
+  const imported = await import(
+    filepath,
+    //@ts-ignore
+    {
+      with: {
+        as: "cachebust",
+        contents: transpiledContent,
+        format: type,
+      } satisfies ImportAttributes,
+    }
```
phryneas, line 72: This will be passed to our custom resolver, the file path will be largely ignored.

sampleWorkspace/localSchema/apollo.config.ts
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  client: {
+    service: {
+      name: "localSchema",
+      localSchemaFile: "./starwarsSchema.graphql",
+    },
+  },
+};
```
phryneas, line 8: We need one "TypeScript, but CommonJS" config for testing, and the rover config seemed like a better candidate for yaml anyways

jerelmiller (approved):
💯

src/language-server/config/cache-busting-resolver.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ *
+ * @param {string} specifier
+ * @param {ResolveContext} context
+ * @param {(specifier: string,context: ResolveContext) => Promise<ResolutionResult>} nextResolve
+ * @returns {Promise<ResolutionResult>}
+ */
+async function resolve(specifier, context, nextResolve) {
+  if (context.importAttributes.as !== "cachebust") {
```
weiatresvu, line 24: It was ⁠`context.importAssertions`. So, in some older versions of Node.js, `context.importAttributes` will be `undefined`. Unfortunately, Cursor is using Node.js 20.9.0 in their latest version.

See, https://nodejs.org/api/module.html#hooks:~:text=v21.0.0%2C%20v20.10.0%2C%20v18.19.0

src/language-server/config/loadTsConfig.ts
```diff
@@ -30,36 +39,39 @@ export const loadTs: Loader = async function loadTs(filepath, content) {
...
+  } catch (error: any) {
+    error.message = `TypeScript Error in ${filepath}:\n${error.message}`;
+    throw error;
   }
+  // eslint-disable-next-line @typescript-eslint/return-await
+  const imported = await import(
+    filepath,
+    //@ts-ignore
+    {
+      with: {
```
yesmeck, line 67: I’m trying to fix #225 and found out the issue is not only related to ⁠`importAttributes`. The ⁠`with` keyword is also newly introduced in new node versions.

See https://nodejs.org/api/esm.html#:~:text=This%20feature%20was%20previously%20named%20%22Import%20assertions%22%2C%20and%20using%20the%20assert%20keyword%20instead%20of%20with.%20Any%20uses%20in%20code%20of%20the%20prior%20assert%20keyword%20should%20be%20updated%20to%20use%20with%20instead.

phryneas: As these are object keys, what happens if you pass this in as an object with both the `with` and `assert` property?

yesmeck: It seems fine to pass both `with` and `assert` to the import function. But I found we won't be able to access the `format` key from `importAssertions`. `importAssertions` only includes two fields despite we setting other keys on the assert object.
CleanShot 2024-10-13 at 15 12 01@2x

phryneas: Hmmm :/ 
Maybe we need to switch to `as: 'cachebust:module'` and `as: 'cachebust:commonjs'` then - or does it maybe work with a different key name than `format`? Maybe that is in some way reserved?

PR #206, opened 2024-09-14
feat: Enable goto definition for upstream language server

trevor-scheer:
We've landed support for the `textDocument/definition` in the language server.

I've tested locally and it seems to work as expected. Is there a place I should add tests for this addition?


phryneas (approved):
Looks good, let's get this in :)

As for tests: we don't have those, since for them to run in CI, we'd need a public rover build. I'll add tests as soon as we have that.

PR #207, opened 2024-09-14
Add script for local build/install

trevor-scheer:
Is something like this interesting? Is there a better way to do it? Might make the PR build redundant?

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "test:extension": "node src/__e2e__/runTests.js",
     "test:textmate": "npx vscode-tmgrammar-snap -s source.ts -g src/language-server/__tests__/fixtures/TypeScript.tmLanguage.json src/language-server/__tests__/fixtures/documents/*.ts",
     "codegen": "graphql-codegen",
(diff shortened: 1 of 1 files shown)


phryneas:
The point of the PR build is more to enable a quick user feedback loop, without having them clone the repo and build the extension themselves.

Locally, I would recommend just running the extension from the Debug Task panel:
image

There might be a slight use case if you really want to test the "built" extension, and not just run it in a folder, but for that I'd probably opt for the CI build as that's just guaranteed to be a lot cleaner than anything from my local installation, which might contain accidentally patched dependencies in `node_modules` etc.

Do you have a use case beyond that that I might be missing? Would still be happy to merge it if we have one :)

trevor-scheer:
I seem to have missed the part where I could open a folder other than the sample workspace 😅
This is definitely unnecessary! I mostly just wanted to try out my changes within my own project.

phryneas:
Yeah, you get a full-fledged VSCode instance there and can open any other folder - it's just pre-opening the sample workspace because we use that for E2E tests :)

PR #209, opened 2024-09-17
Change default recommended configuration file format to `json`.

phryneas:
This changes the default recommended configuration file format to `.json`.

I would prefer to make this `yaml` instead, but the configuration's JSON schema is natively supported  by VSCode for `.json` files and needs an additional extension to be installed by the user to support `.yaml` files, so I believe going this way will result in less onboarding friction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,9 @@ To link a project to a published schema, edit the `apollo.config.json` file to l
 ```json
 {
   "client": {
-    "service": "graphos-graph-id",
-  },
-};
+    "service": "graphos-graph-id"
+  }
+}
 ```
 
 The `service` name is the name of the graph you've created in [GraphOS Studio](https://studio.apollographql.com).
@@ -89,10 +89,10 @@ Sometimes it may make sense to link the editor to a locally running version of a
   "client": {
     "service": {
       "name": "my-graphql-app",
-      "url": "http://localhost:4000/graphql",
-    },
-  },
-};
+      "url": "http://localhost:4000/graphql"
+    }
+  }
+}
 ```
 
 Linking to the local schema won't provide all features, such as switching graph variants and performance metrics.
@@ -113,10 +113,10 @@ To link to a local schema file, add the following to the `apollo.config.json` fi
   "client": {
     "service": {
       // can be a string pointing to a single file or an array of strings
-      "localSchemaFile": "./path/to/schema.graphql",
-    },
-  },
-};
+      "localSchemaFile": "./path/to/schema.graphql"
+    }
+  }
+}
 ```
 
 </details>
@@ -134,9 +134,9 @@ Client side schema definitions can be spread throughout the client app project a
     // array of glob patterns
     "includes": ["./src/**/*.js"],
     // array of glob patterns
-    "excludes": ["**/__tests__/**"],
-  },
-};
+    "excludes": ["**/__tests__/**"]
+  }
+}
 ```
 
 <h3 id="get-the-extension">Get the extension</h3>
@@ -218,11 +218,11 @@ _Optional_ - custom tagged template literal.
 
 When using GraphQL with JavaScript or TypeScript projects, it is common to use the `gql` tagged template literal to write out operations. Apollo tools look through your files for the `gql` tag to extract your queries, so if you use a different template literal, you can configure it like so:
 
-```js
-module.exports = {
-  client: {
-    tagName: "graphql", // highlight-line
-    service: ...
+```json
+{
+  "client": {
+    "tagName": "graphql",
+    "service": //...
   }
-};
+}
 ```
--- a/README.md
+++ b/README.md
@@ -35,22 +35,17 @@ For the contents of this configuration file, select one of these options:
 
 <details>
 <summary>
-<h3>Setting the Extension up for use with Apollo GraphOS</h3>
+<h3>Configure extension for schemas published to Apollo GraphOS</h3>
 </summary>
 
-To get all the benefits of the VS Code experience, it's best to link the schema that is being developed against before installing the extension. The best way to do that is by [publishing a schema](https://www.apollographql.com/docs/graphos/delivery/publishing-schemas/) to the Apollo schema registry. After that's done:
+To get all the benefits of the VS Code experience, it's best to link the schema that is being developed against before installing the extension. The best way to do that is by [publishing a schema](https://www.apollographql.com/docs/graphos/delivery/publishing-schemas/) to the Apollo schema registry.
 
-1. Create an `apollo.config.json` file at the root of the project.
-2. Obtain a [Personal API key](https://www.apollographql.com/docs/graphos/api-keys) from GraphOS Studio.
-
-<h3 id="apollo-config">Setting up an Apollo config</h3>
-
-To link a project to a published schema, edit the `apollo.config.json` file to look like this:
+After that's done, edit the `apollo.config.json` file to look like this:
 
 ```json
 {
   "client": {
-    "service": "graphos-graph-id"
+    "service": "graphos-graph-name"
   }
 }
 ```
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -26,20 +26,31 @@ The Apollo GraphQL extension for VS Code brings an all-in-one tooling experience
 
 <h2 id="getting-started">Getting started</h2>
 
+For the VS Code plugin to know how to find the schema, it needs to be linked to either a published schema or a local one.
```
phryneas, line 29: @Meschreiber, could I ask you to take a quick look at these changes? Once we ship it here and the `details` tags are working, I would add the same changes to the corresponding Docs page.

I made these changes as we get a lot of reviews that complain that you *need* a GraphOS account, which is not true. It seems the alternative configurations for local service and local file were not prominent enough until now.

By making these sections collapsible, I'm trying to make all of them more discoverable.

src/build.js
```diff
@@ -89,6 +90,19 @@ const buildJsonSchemaPlugin = {
...
 };
 
+const resolvePlugin = {
+  name: "resolve",
+  setup(build) {
+    console.log("setup");
+    build.onResolve(
+      { filter: /^jsonc-parser$/ },
+      async ({ path, ...options }) => {
+        return build.resolve("jsonc-parser/lib/esm/main.js", options);
```
phryneas, line 100: This package offers ESM and UMD. If I import the UMD in VSCode (the default), it makes the extension crash - if I import the ESM in the tests, the tests crash.
This way, the tests get UMD and the extension bundles with the ESM import.

README.md
```diff
@@ -195,11 +218,11 @@ _Optional_ - custom tagged template literal.
 
 When using GraphQL with JavaScript or TypeScript projects, it is common to use the `gql` tagged template literal to write out operations. Apollo tools look through your files for the `gql` tag to extract your queries, so if you use a different template literal, you can configure it like so:
 
-```js
-module.exports = {
-  client: {
-    tagName: "graphql", // highlight-line
-    service: ...
```
phryneas, line 202: image
This actually rendered out in the [Marketplace](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo), so I'm removing it

README.md
Meschreiber: Thanks for the ping here!

I didn't realize the README for this project had a lot of duplicated content with the docs page. Did we consider just linking to the docs page rather than maintaining both places?/Do you think the benefit of visibility (particularly in the VSC marketplace) outweighs the minor burden of maintaining in both places?

README.md
```diff
@@ -26,22 +26,33 @@ The Apollo GraphQL extension for VS Code brings an all-in-one tooling experience
...
+<details>
+<summary>
+<h3>Setting the Extension up for use with Apollo GraphOS</h3>
+</summary>
+
 To get all the benefits of the VS Code experience, it's best to link the schema that is being developed against before installing the extension. The best way to do that is by [publishing a schema](https://www.apollographql.com/docs/graphos/delivery/publishing-schemas/) to the Apollo schema registry. After that's done:
 
-1. Create an `apollo.config.js` file at the root of the project.
-   Alternatively, you can create a `cjs`, `mjs`, or `ts` file with the same configuration.
+1. Create an `apollo.config.json` file at the root of the project.
```
Meschreiber: Seems like  this step is no longer necessary since it's mentioned above?

README.md
```diff
@@ -26,22 +26,33 @@ The Apollo GraphQL extension for VS Code brings an all-in-one tooling experience
...
+For the VS Code plugin to know how to find the schema, it needs to be linked to either a published schema or a local one.
+
+First, create an `apollo.config.json` file at the root of the project.
+Alternatively, you can create a `yaml`, `cjs`, `mjs`, or `ts` file with the same configuration.
+
+For the contents of this configuration file, select one of these options:
+
+<details>
+<summary>
+<h3>Setting the Extension up for use with Apollo GraphOS</h3>
```
Meschreiber: I think it's important that the title clearly states this is the "published" option.

```suggestion
<h3>Configure extension for schemas published to Apollo GraphOS</h3>
```

README.md
```diff
@@ -26,22 +26,33 @@ The Apollo GraphQL extension for VS Code brings an all-in-one tooling experience
...
-  client: {
-    service: "my-graphql-app",
-  },
-};
+```json
+{
+  "client": {
+    "service": "graphos-graph-id"
+  }
+}
```
Meschreiber: I would remove this and the following `<h3>` and just have one continuous set of instructions. E.g.,

1. Obtain a personal API key.
2. Edit the `apollo.config.json` to look like this...
3. To pull down your schema, create an `.env` file ...

README.md
```diff
@@ -50,7 +61,7 @@ See [additional configuration options](#additional-apollo-config-options).
 
 <h3 id="api-key">Setting up the <code>.env</code> file</h3>
 
-To authenticate with GraphOS Studio to pull down your schema, create a `.env` file in the same directory as the `apollo.config.js` file. This should be an untracked file (that is, don't commit it to Git).
+To authenticate with GraphOS Studio to pull down your schema, create a `.env` file in the same directory as the `apollo.config.json` file. This should be an untracked file (that is, don't commit it to Git).
 
 Then go to your [User Settings page](https://studio.apollographql.com/user-settings/api-keys?referrer=docs-content) in GraphOS Studio to create a new Personal API key.
```
Meschreiber, line 59: Is this different than the API key they needed to previously obtain?

README.md
```diff
@@ -64,56 +75,68 @@ APOLLO_KEY=<enter copied key here>
 
 After this is done, VS Code can be reloaded and the Apollo integration will connect to GraphOS Studio to provide autocomplete, validation, and more.
 
-<h3 id="local-schemas">Local schemas</h3>
+</details>
+
+<details>
+<summary>
+<h3 id="local-schemas">Setting the Extension up for use with a local schema</h3>
```
Meschreiber:
```suggestion
<h3 id="local-schemas">Configure extension for local schemas</h3>
```

README.md
```diff
@@ -64,56 +75,68 @@ APOLLO_KEY=<enter copied key here>
...
+}
 ```
 
 Linking to the local schema won't provide all features, such as switching graph variants and performance metrics.
 
-<h3 id="local-schema-files">Local schema files</h3>
+</details>
+
+<details>
+<summary>
```
Meschreiber, line 96: I would include this under the same header (Configure extension for local schemas) and present the two options has `<h4>`s.

- Use introspection of a locally running service
- Use with a local schema file.

IMO, since these are two flavors of the "Configure extension for local schemas" they should be separate `<h4>`s within that collapsible section.

Meschreiber (approved):
Just some structural suggestions and a question. Otherwise LGTM!

src/build.js
```diff
@@ -89,6 +90,19 @@ const buildJsonSchemaPlugin = {
   },
 };
 
+const resolvePlugin = {
+  name: "resolve",
+  setup(build) {
+    console.log("setup");
```
jerelmiller:
```suggestion
```

Is this needed?

jerelmiller (approved):
Spotted one stray log, but other than Maria's requests, I've got nothing else.

README.md
phryneas: Yes, definitely - especially the screenshots help a lot showcasing the features we have. I'm happy to keep both in sync :)

src/build.js
phryneas: Not anymore, good catch :D

README.md
phryneas: Yeah, and this is only a summary of what's coming next, so I'll delete the whole thing.

README.md
phryneas: No, the exact same one. This whole thing is just a bit too repetitive :sweat_smile:
Removed the earlier mention.

README.md
phryneas: People have been missing these options a lot, so I want to have both options as visible as the GraphOS one. It's really pulling down our marketplace ratings and leads to bad reviews from non-GraphOS users (which make up a *big* part of our userbase - we have 500k downloads).

Meschreiber (approved):
Thanks for restructuring, looks a lot clearer!

PR #210, opened 2024-09-18
Add upstream LSP tests to CI

phryneas:


phryneas:
As this just adds tests I'm gonna merge without review.

Issue #187, opened 2024-08-29
Unable to configure apollo extension


mklemarczyk:
Hello, I got the last update 2.2.0 and it does not work anymore for my project.

I use the vs code workspace and my apollo config file is in the root folder. I use nx workspace folder structure fo angular project, that means the folder structure is quite deep.

I can provide more details later in the afternoon. Error is written to the console that it does not find the configuration file, even if you open the configuration file. I couldn’t find any helpful information in the verbose logs of the extension.

Last extension version that works for me is 1.20.0. I use apollo client npm in version 3.11.2 and apollo-angular in version 7.0.2.

Any version of extension 2.x has the same issue. I seen that your extension has very low reputation at VS Code Marketplace.

phryneas:
Hi Maciej,

it would be great if you could share the name & contents of your configuration file (of course without sensitive details), an idea of your folder layout (where is the config file, which folder do you have open), and the exact error message.
I'll look into it as soon as I have more information :)

bhpark1013:
```
module.exports = {
  client: {
    service: {
      localSchemaFile: './schema.gql',
    },
  },
  exclude: ['**/*.e2e-spec.ts'],
};

```
I am also having similar experience. I haven't try earlier version cause i have some files to exclude(which seems not working option on version 1.x). I am trying to link with local schema without using Apollo schema registry. 

I have logs which seems not that helpful
```
.vscode/extensions/apollographql.vscode-apollo-2.2.0/lib/language-server/server.js:1035
`)}t(Swn,"dedentBlockStringValue");function Ewn(a){le  //...and so on
[Error - 6:14:03 PM] Server process exited with code 1.
[Error - 6:14:03 PM] Client Apollo GraphQL: connection to server is erroring.
Channel closed
```
my apollo/server version is: "@apollo/server": 4.9.5,

Can i get any help?

phryneas:
I'm going to do a bit of digging what code that points to.
Meanwhile: is this already from the "Apollo" output or is there maybe additional information there?
![image](https://github.com/user-attachments/assets/da8852da-4681-4a2a-a76d-7919c1d95f5c)

phryneas:
@bhpark1013 also, could I please get the full unshortened error with the full stack trace? That might actually tell me something :)

phryneas:
[...]

PR #211, opened 2024-09-19
Avoid detection if .js config file is ESM or CommonJs, just try both.

phryneas:
This came up here: https://github.com/apollographql/vscode-graphql/issues/187#issuecomment-2358950870

Because we relied on the original `resolve` call to determine if a file was a module or commonjs, that call would apply the normal ESM detection logic - if your `package.json` had `type: 'module'`, this means that your config file could not be CommonJS anymore.

This approach now just tries to load it as ESM and if that fails tries again with CJS - we do the same for TypeScript already.

### DIFF
--- a/sampleWorkspace/configFileTypes/jsConfigWithESM/package.json
+++ b/sampleWorkspace/configFileTypes/jsConfigWithESM/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "test",
+  "type": "commonjs"
+}
--- a/src/language-server/config/cache-busting-resolver.js
+++ b/src/language-server/config/cache-busting-resolver.js
@@ -24,20 +24,12 @@ async function resolve(specifier, context, nextResolve) {
   if (context.importAttributes.as !== "cachebust") {
     return nextResolve(specifier, context);
   }
-  if (context.importAttributes.format) {
-    // no need to resolve at all, we have all necessary information
-    return {
-      url: bustFileName(specifier),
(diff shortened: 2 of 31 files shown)


src/language-server/config/cache-busting-resolver.js
```diff
@@ -24,20 +24,12 @@ async function resolve(specifier, context, nextResolve) {
...
-  }
-  const result = await nextResolve(specifier, context);
+  // no need to resolve at all, we have all necessary information
   return {
-    ...result,
-    url: bustFileName(result.url),
+    url: bustFileName(specifier),
+    format: context.importAttributes.format,
     importAttributes: context.importAttributes,
+    shortCircuit: true,
```
phryneas, line 32: We don't trust the original detection algorithm at all anymore, `format` is required now and we just pass this in from the outside.

sampleWorkspace/configFileTypes/jsConfigWithESM/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "test",
+  "type": "commonjs"
```
phryneas, line 3: Deliberately specifying the opposite here - the config file is `.js` with ESM contents, the package here has `type: commonjs`, we still want it to work.

The E2E tests will try all possible combinations on top.

jerelmiller (approved):
Seems reasonable to me!

PR #214, opened 2024-09-19
adjust Rover debug logging to use tracing level

phryneas:
Some clean up in preparation on a release. 
We don't need a `DEBUG` constant, there is already a "tracing" level built into VSCode, and it's enabled for the extension via the setting in 
https://github.com/apollographql/vscode-graphql/blob/5449afbe7e4835630efbdfd9555d85a8c9888a87/package.json#L118-L128

### DIFF
--- a/src/language-server/project/rover/project.ts
+++ b/src/language-server/project/rover/project.ts
@@ -36,8 +36,8 @@ import { VSCodeConnection } from "../../server";
 import { getLanguageIdForExtension } from "../../utilities/languageIdForExtension";
 import { extname } from "node:path";
 import type { FileExtension } from "../../../tools/utilities/languageInformation";
-
-export const DEBUG = true;
+import { Debug } from "../../utilities";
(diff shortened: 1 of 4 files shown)


src/language-server/project/rover/project.ts
```diff
@@ -115,11 +115,14 @@ export class RoverProject extends GraphQLProject {
...
-      console.log("sending notification %o", {
+    Debug.traceMessage(
+      "[Rover] Sending notification: " + type.method,
+      "[Rover] Sending notification %o",
+      {
         type: type.method,
         params,
-      });
+      },
+    );
```
phryneas, line 125: Signature here is
```
Debug.traceMessage(
  short,
  verbose = short,
  ...verboseFormatParams
)
```

where `short` will be logged for log level "messages" and `verbose` for log level "verbose".

jerelmiller (approved):
💯

PR #215, opened 2024-09-19
remove feature flag around Rover projects

phryneas:
README changes are separate in #220 now, this PR does contain the screenshot for the README though, so that #220 can already point at `main`.

Note that this PR *does* contain the JSON schema that semi-documents the `rover` configuration key. I'll verify if we want to keep that here or move it over as well.


jerelmiller (approved):
💯

Meschreiber (approved):
Thought I might be able to directly commit for copyedits, but it looks like that's not the case. Sending a copyedited version of the README over Slack!

phryneas:
Added Maria's Copyedit - thank you @Meschreiber!

PR #216, opened 2024-09-20
add json schemas for supergraph.yaml

phryneas:
Following up on our discussion from last week.


package.json
```diff
@@ -257,6 +257,15 @@
       {
         "fileMatch": "apollo.config.yaml",
         "url": "./schemas/apollo.config.schema.json"
+      },
+      {
+        "fileMatch": "apollo-router.yaml",
+        "_url": "https://github.com/apollographql/router-template/raw/main/.apollo/router_config_schema.json",
+        "url": "./schemas/router_config_schema.json"
```
phryneas, line 264: I'm not sure if I should link to the schema on GitHub or a local copy here. Any thoughts @dylan-apollo?

phryneas: I'm not sure if I should link to the schema on GitHub or a local copy here. Any thoughts @dylan-apollo?
```suggestion
        "url": "https://github.com/apollographql/router-template/raw/main/.apollo/router_config_schema.json"
```

schemas/router_config_schema.json
```diff
@@ -0,0 +1,8342 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Configuration",
+  "description": "The configuration for the router.\n\nCan be created through `serde::Deserialize` from various formats, or inline in Rust code with `serde_json::json!` and `serde_json::from_value`.",
```
phryneas: This description is a bit unfortunate as it shows up in the editor:

image

jerelmiller (approved):

schemas/router_config_schema.json
dylan-apollo: Yeah, this is something we should run by the router team. The point of this schema is to get editor support, so we should be updating our internal docs to be more useful for external consumers

package.json
dylan-apollo: If you do a local schema, you should also set up Renovate to regenerate it when new router releases come out—which would mean releasing a new version of the extension every week or two with the latest.

Or, I suppose, if we're keeping it up to date, we could also point the URL to _this_ repo instead of `router-template`?

dylan-apollo:
Something I didn't consider, that we should test, is how this works with unstable router features. We have new config for Router 2.0, I wonder if that shows up as an error, with this change 😐

package.json
phryneas: We'll already have releases trickling in other places - Apollo Client release will trigger Apollo Devtools release, which will long-term trigger a release of the VSCode Devtools (or at least a PR).

I'm not too keep on doing that repeatedly with even more upstream sources, so I'd honestly rather point at another repo that is kept kinda up-to-date.

phryneas:
> Something I didn't consider, that we should test, is how this works with unstable router features. We have new config for Router 2.0, I wonder if that shows up as an error, with this change 😐

Two strategies come to mind:

* Have some "config-version" key in there and both versions in the schema as alternative schemas (maybe this could even reference two different urls? I'm not that deep in json schemas yet)
* We add another config for an `apollo-router-2.yaml` or similar.

package.json
phryneas: Oh, we could also consider publishing to https://www.schemastore.org/json/ and linking to that?

See https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md

dylan-apollo:
@jerelmiller can we keep/merge the supergraph schema one (which pretty much never changes and should be safe) and just take out the Router config stuff for now? I don't think I'll have time to fix the red squigglies on connectors features before we want this released.

phryneas:
@dylan-apollo I've reduced the scope to `supergraph.yaml` for now - should I get this in?

dylan-apollo (approved):
Let's do it, thanks!!

PR #226, opened 2024-10-11
fix importAttributes is not avaliable in old node versions

### LINKED ISSUE #225: `importAttributes` not available in old node versions
reported by @weiatresvu here https://github.com/apollographql/vscode-graphql/pull/204#pullrequestreview-2361756806

> It was ⁠context.importAssertions. So, in some older versions of Node.js, context importAttributes will be undefined. Unfortunately, Cursor is using Node.js 20.9.0 in their latest version.
> 
> See, [nodejs.org/api/module.html#hooks:~:text=v21.0.0%2C v20.10.0%2C v18.19.0](https://nodejs.org/api/module.html#hooks:~:text=v21.0.0%2C%20v20.10.0%2C%20v18.19.0)

This would also impact VSCode before 1.92 (see [chart](https://github.com/ewanharris/vscode-versions), our current minimum is 1.90) so we will need a workaround for that.

@weiatresvu would you be open for a PR on this? :)

yesmeck:
fix #225 

Tested with VSCode 1.91.1 (Node.js 20.9.0) and VSCode 1.94.0 (Node.js 20.16.0).

### DIFF
--- a/src/language-server/config/cache-busting-resolver.js
+++ b/src/language-server/config/cache-busting-resolver.js
@@ -5,7 +5,7 @@ const { pathToFileURL } = require("node:url");
 
 /**
  * importAssertions was renamed to importAttributes after following versions of Node.js.
- * Once we hit a minimum of 1.92, we can remove the legacy check and
+ * Once we hit a minimum of v1.92 of VSCode, we can remove the legacy check and
  * use `importAttributes` directly.
  *
  * - v21.0.0
--- a/src/language-server/config/cache-busting-resolver.js
+++ b/src/language-server/config/cache-busting-resolver.js
@@ -1,7 +1,6 @@
 // @ts-check
 const { pathToFileURL } = require("node:url");
 
-/** @import { ResolveContext, ResolutionResult, LoadResult, ImportContext } from "./cache-busting-resolver.types" */
 /** @import { ResolveContext, ResolutionResult, LoadResult, ImportContext,ImportAttributes } from "./cache-busting-resolver.types" */
 /**
  * importAssertions was renamed to importAttributes in newer versions of Node.js.
--- a/src/language-server/config/cache-busting-resolver.js
+++ b/src/language-server/config/cache-busting-resolver.js
@@ -1,7 +1,7 @@
 // @ts-check
 const { pathToFileURL } = require("node:url");
 
-/** @import { ResolveContext, ResolutionResult, LoadResult, ImportContext,ImportAttributes } from "./cache-busting-resolver.types" */
+/** @import { ResolveContext, ResolutionResult, LoadResult, ImportContext, ImportAttributes } from "./cache-busting-resolver.types" */
 /**
  * importAssertions was renamed to importAttributes in newer versions of Node.js.
  *
--- a/src/language-server/config/cache-busting-resolver.js
+++ b/src/language-server/config/cache-busting-resolver.js
@@ -106,7 +106,7 @@ async function load(url, context, nextLoad) {
   const contents =
     "contents" in importAttributes
       ? importAttributes.contents
-      : Object.keys(importAttributes)[1];
+      : Object.keys(importAttributes).find((key) => key != "as");
   return {
     format,
     shortCircuit: true,
--- a/src/language-server/config/cache-busting-resolver.js
+++ b/src/language-server/config/cache-busting-resolver.js
@@ -9,7 +9,6 @@ const { pathToFileURL } = require("node:url");
  * @param {ResolveContext|ImportContext} context
  * @returns {boolean}
  */
-
 function isImportAttributesAvailable(context) {
   return "importAttributes" in context;
 }
--- a/src/language-server/config/cache-busting-resolver.js
+++ b/src/language-server/config/cache-busting-resolver.js
@@ -7,13 +7,37 @@ const { pathToFileURL } = require("node:url");
  * importAssertions was renamed to importAttributes in newer versions of Node.js.
  *
  * @param {ResolveContext|ImportContext} context
+ * @returns {boolean}
+ */
+
+function isImportAttributesAvailable(context) {
+  return "importAttributes" in context;
+}
+
+/**
+ * importAssertions was renamed to importAttributes in newer versions of Node.js.
+ *
+ * @param {ResolveContext|ImportContext} context
  * @returns {"importAttributes"|"importAssertions"}
  */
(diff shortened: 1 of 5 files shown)


phryneas:
I'm merging in `main`, as I've added a CI run for 1.90.0 there.

src/language-server/config/cache-busting-resolver.js
```diff
@@ -41,13 +80,19 @@ async function resolve(specifier, context, nextResolve) {
...
-  if (context.importAttributes.as !== "cachebust") {
+  const importAttributes = resolveImportAttributes(context);
+  const [as, format] = importAttributes.as.split(":");
+  if (as !== "cachebust") {
     return nextLoad(url, context);
   }
+  const contents =
+    "contents" in importAttributes
+      ? importAttributes.contents
+      : Object.keys(importAttributes)[1];
```
yesmeck: It’s pretty weird here is that the second key of importAssertions is the content of the module.

phryneas: Yes, seriously. But if it works it is what it is - this only applies to a small version range and will go away at some point in the future 🤷

src/language-server/config/cache-busting-resolver.js
```diff
@@ -3,6 +3,38 @@ const { pathToFileURL } = require("node:url");
 
 /** @import { ResolveContext, ResolutionResult, LoadResult, ImportContext } from "./cache-busting-resolver.types" */
```
phryneas, line 5:
```suggestion
/** @import { ResolveContext, ResolutionResult, LoadResult, ImportContext,ImportAttributes } from "./cache-busting-resolver.types" */
```

src/language-server/config/cache-busting-resolver.js
```diff
@@ -1,7 +1,38 @@
 // @ts-check
 const { pathToFileURL } = require("node:url");
 
-/** @import { ResolveContext, ResolutionResult, LoadResult, ImportContext } from "./cache-busting-resolver.types" */
+/** @import { ResolveContext, ResolutionResult, LoadResult, ImportContext, ImportAttributes } from "./cache-busting-resolver.types" */
+/**
+ * importAssertions was renamed to importAttributes in newer versions of Node.js.
+ *
+ * @param {ResolveContext|ImportContext} context
+ * @returns {boolean}
```
phryneas:
```suggestion
 * @returns {context is {importAttributes: ImportAttributes}}
```

src/language-server/config/cache-busting-resolver.js
```diff
@@ -21,14 +52,16 @@ function bustFileName(specifier) {
  * @returns {Promise<ResolutionResult>}
  */
 async function resolve(specifier, context, nextResolve) {
-  if (context.importAttributes.as !== "cachebust") {
+  const importAttributes = resolveImportAttributes(context);
+  const [as, format] = importAttributes.as.split(":");
```
phryneas: This might need to move out into another function to make TS happy here

phryneas:
Generally, could you please add some comment to document that once we hit a minimum of 1.92, most of these workarounds can be removed again, and reference the issue you started?
So we don't confuse future maintainers :)

src/language-server/config/cache-busting-resolver.js
```diff
@@ -21,14 +52,16 @@ function bustFileName(specifier) {
...
+  if (as !== "cachebust") {
     return nextResolve(specifier, context);
   }
```
phryneas:
```suggestion
  const importAttributes = resolveImportAttributes(context);
  if (!importAttributes.as) {
    return nextResolve(specifier, context);
  }
  const [as, format] = importAttributes.as.split(":");
  if (as !== "cachebust" || (format !== "module" && format !== "commonjs")) {
    return nextResolve(specifier, context);
  }
```

src/language-server/config/cache-busting-resolver.js
```diff
@@ -41,13 +74,19 @@ async function resolve(specifier, context, nextResolve) {
  * @returns {Promise<LoadResult>}
  */
 async function load(url, context, nextLoad) {
-  if (context.importAttributes.as !== "cachebust") {
+  const importAttributes = resolveImportAttributes(context);
+  const [as, format] = importAttributes.as.split(":");
+  if (as !== "cachebust") {
     return nextLoad(url, context);
```
phryneas:
```suggestion
  const importAttributes = resolveImportAttributes(context);
  if (!importAttributes.as) {
    return nextLoad(url, context);
  }
  const [as, format] = importAttributes.as.split(":");
  if (as !== "cachebust" || (format !== "module" && format !== "commonjs")) {
    return nextLoad(url, context);
```

phryneas: I think these two changes should make TS happy here

src/language-server/config/cache-busting-resolver.js
```diff
@@ -53,8 +77,8 @@ function bustFileName(specifier) {
  */
 async function resolve(specifier, context, nextResolve) {
   const importAttributes = resolveImportAttributes(context);
-  const [as, format] = importAttributes.as.split(":");
-  if (as !== "cachebust") {
+  const format = resolveConfigFormat(importAttributes);
```
phryneas, line 80: I like this a lot better than my suggestion :)

src/language-server/config/cache-busting-resolver.js
yesmeck: I've moved it out to a function.

phryneas:
Okay, once the tests come back happy all that's left is a changeset - could you please run `npx changeset` and add a changeset for a patch-level change? :)

phryneas:
Huh, that test runs fine for me locally. I'll investigate.

yesmeck:
It seems we can't set `contents` on `importAssertions`. It's so weird that it was working for me.

phryneas:
This is just super weird :(
Any more ideas?
Putting `contents` into `as` too? 😢

yesmeck:
Can you rerun CI? It works for me locally now. We need to set both `⁠format` and `⁠contents` on the assert object, even though we can’t access the format from ⁠importAssertions, and the order of ⁠format and ⁠contents as keys also matters.

phryneas (approved):
And we're green.
Thank you *very* much for your hard work with this!

phryneas:
Released as https://github.com/apollographql/vscode-graphql/releases/tag/v2.3.5

PR #228, opened 2024-10-15
fix ci & run ci with multiple VSCode versions

phryneas:


phryneas:
Based on #225 and #226, this is the expected result. Merging.

PR #230, opened 2024-10-18
Add support for Apollo Connectors Mapping Syntax

phryneas:


syntaxes/graphql.connectors.json
```diff
@@ -0,0 +1,105 @@
...
+            "2": { "name": "punctuation.definition.string.begin.graphql" }
+          },
+          "endCaptures": {
+            "1": { "name": "string.quoted.double.graphql" },
+            "2": { "name": "punctuation.definition.string.end.graphql" },
+            "3": { "name": "invalid.illegal.newline.graphql" }
+          },
+          "patterns": [
+            { "include": "source.graphql#graphql-field" },
+            { "include": "source.graphql#graphql-comma" }
```
phryneas: This is the point to specify extra rules for the real grammar

syntaxes/graphql.connectors.json
```diff
@@ -0,0 +1,68 @@
...
+      "beginCaptures": {
+        "1": { "name": "variable.parameter.graphql" },
+        "2": { "name": "punctuation.colon.graphql" }
+      },
+      "endCaptures": {
+        "3": { "name": "punctuation.comma.graphql" }
+      },
+      "patterns": [
+        {
+          "contentName": "meta.embedded.block.jsonselection.graphql",
```
phryneas: Right now I'm stuck at the point where I'd need to inject the JSONSelection grammar into here.

syntaxes/connectors.jsonselection.yaml
```diff
@@ -0,0 +1,270 @@
+# use 'js-yaml' to convert to json
+# npx js-yaml syntaxes/connectors.jsonselection.yaml >| syntaxes/connectors.jsonselection.json
+name: "Apollo Connectors JSONSelection"
```
dylan-apollo: Is this user-facing at all? We've avoided using this name anywhere user-facing for now, since we're going to change it, we just haven't decided to what yet 😅

syntaxes/connectors.jsonselection.yaml
```diff
@@ -0,0 +1,270 @@
...
+      - include: "#PathWithSubSelection"
+      - include: "#NamedFieldSelection"
+      - include: "#NamedPathSelection"
+      - include: "#NamedGroupSelection"
+  # NamedPathSelection   ::= Alias PathSelection
+  NamedPathSelection:
+    begin: "\\b([a-zA-Z][0-9a-zA-Z_]*)\\s*(:)"
+    beginCaptures:
+      "1":
+        name: "string.unquoted.NamedPathSelection.alias"
```
dylan-apollo: So can each of these sections get their own colors? I'd love to have one color for the "output" name (alias in this case) and another for things that are inputs.

syntaxes/graphql.connectors.json
```diff
@@ -0,0 +1,108 @@
...
+            { "include": "source.graphql#graphql-skip-newlines" }
+          ]
+        },
+        { "include": "source.graphql#graphql-argument" },
+        { "include": "source.graphql#literal-quasi-embedded" }
+      ]
+    },
+    "connect-jsonselection-argument": {
+      "name": "meta.argument.connect.jsonselection.graphql",
+      "begin": "\\s*(selection|body)(?:\\s*(:))",
```
dylan-apollo: Are these recursive or only top-level? `body` is within `http`, is the only reason I ask

syntaxes/graphql.connectors.json
```diff
@@ -0,0 +1,108 @@
...
+            }
+          ]
+        },
+        { "include": "source.graphql#graphql-skip-newlines" }
+      ]
+    },
+    "connect-string-jsonselection-argument": {
+      "name": "meta.argument.connect.string.jsonselection.graphql",
+      "contentName": "meta.embedded.block.string.jsonselection.graphql",
+      "begin": "\\s*(GET|POST)(?:\\s*(:))",
```
dylan-apollo: Oh I see, is this what is checking the sub-argument to find `body`? We also support `PUT`, `PATCH`, and `DELETE`

dylan-apollo (approved):
I'll be honest, I don't really understand how this works, but your demo looked great! Added a few questions/comments

syntaxes/graphql.connectors.json
phryneas: Semi-rescursive. If you look a few lines up you'll see that I have an inlined pattern in `#connect-arguments` that looks for `\\s*(http)(?:\\s*(:))` and then applies `#connect-arguments` to that recursively.

Although I think this might actually work more on accident... I'll give this another read or two :D

syntaxes/graphql.connectors.json
phryneas: This is to find usages where we don't do a full jsonselection "selection set", but a string that contains selections between `{}`.

syntaxes/connectors.jsonselection.yaml
phryneas: That's a good call-out - the only "user-facing" thing right now is that we introdude a new language to VSCode:

https://github.com/apollographql/vscode-graphql/blob/0ea12de909a837847cdedc704c80f4278f3ec879/package.json#L159-L165

That means it will pop up if you edit a file called `test.apollo.jsonselection`, but also if you use the "change file language" dropdown:

image


We might want to give that language a better user-facing name.

It's technically also visible if you inspect syntax highlighting tokens, but I would not really consider that "user-facing":

image


What should I better name this instead? `apollo.connectors.syntax`?

syntaxes/connectors.jsonselection.yaml
phryneas: Yes, they can!

Open a text editor and execute the "Developer: inspect editor tokens and scopes" command

image

that will give you this pop-up:

image


And then you can move your cursor around and see what tokens things in other languages have, as an inspiration.

---

image
image

You see that GraphQL uses `string.unquoted` and `variable` as highlighting scopes here, but I also see that the `alias` is never applied because `variable.KeyPath.key.identifier` seems to have a higher priority somehow.

I'll debug that :)

phryneas: So... with some changes:

image

syntaxes/graphql.connectors.json
phryneas: Also, added support for PUT|PATCH|DELETE ✅

syntaxes/graphql.connectors.json
phryneas: Should be more solid now, I added a `#connect-jsonselection-nested-argument` pattern

syntaxes/connectors.jsonselection.yaml
dylan-apollo: Amazing! A couple more notes on what we might want to add in future iterations:

1. `$` and `@` as well as an identifier following $, like `$args`, `$this`, etc. are sort of our keywords? They're special, at very least, and might be worth highlighting separately.
2. Anything in `$()` or as an argument to a method is sort of a different language mode. We call them literals, but they can also have some expressions within them. So `${ blah: $.blah }` and `$({ blah: $.blah })` are both valid, and look very similar, but do different things. It'd be nice to highlight that difference somehow (effectively, just the presence of the parens causes it).

syntaxes/connectors.jsonselection.yaml
dylan-apollo: Yeah, that sounds generic enough to be okay. We also use the term "mapping" a lot when referring to it right now, like "Apollo Connectors Mapping Syntax", though that's really verbose.

dylan-apollo (approved):

syntaxes/connectors.jsonselection.yaml
phryneas: I can configure `{(` and `)}` as a special kind of bracket pairs (this is VSCode, not TextMate!) so they will always have the same color and highlight in pairs together:
image

lennyburdette: that's cool! i don't think it's exactly what we want tho. the way i see it, when we're inside parens (for both literal values like `$(...)` or for method arguments like `->method(...)`), we're now in "literal value" mode instead of "selection" mode. So I'd color the outer parens as one thing, and any inner values to indicate that they're different from selections. i dunno exactly what that looks like

phryneas: As for 1., does this work for you @dylan-apollo ?
image

phryneas: @lennyburdette then I'm undoing that change and we go back to this:
image


Honestly I don't know if I can do a lot to visually highlight that difference more, apart from adding a full other "sub-language", which I'd rather avoid as this would become even more unmaintainable 😅

dylan-apollo: Maybe we just color parens differently than `{}` for now, since parens always represent entering literal mode? We could do something special with semantic tokens or something if we really want in the future

dylan-apollo: The `@` and `$` stuff looks great!

phryneas: > Maybe we just color parens differently than {} for now, since parens always represent entering literal mode? We could do something special with semantic tokens or something if we really want in the future

We'll need to test that out in the non-VSCode editors. VSCode pretty much ignores everything else and applies its "bracket pair coloring" logic over the top of configured bracket pairs (which we want to have so it highlights start and end).

dylan-apollo: Then maybe it's just fine 😅. Certainly shouldn't block getting the highlighting of the rest of the language out there

syntaxes/connectors.jsonselection.yaml
phryneas: image

Renamed it, and also put it into `[]` because that will sort it alphabetically at the bottom of the list. There's no need for this implementation detail to be the first thing on the list every time people open it :)

syntaxes/graphql.json
```diff
@@ -342,7 +342,14 @@
...
     "graphql-directive": {
-      "begin": "\\s*((@)\\s*([_A-Za-z][_0-9A-Za-z]*))",
+      "contentName": "meta.embedded.block.directive.graphql",
+      "begin": "\\s*(?=@)",
+      "end": "(?=.)",
+      "applyEndPatternLast": 1,
+      "patterns": [{ "include": "#graphql-directive-body" }]
+    },
+    "graphql-directive-body": {
+      "begin": "((@)\\s*([_A-Za-z][_0-9A-Za-z]*))",
```
phryneas, line 352: I managed to get this down to a minimum amount of changes, but unfortunately not down to none.

syntaxes/graphql.connectors.yaml
```diff
@@ -0,0 +1,167 @@
+scopeName: source.apollo.connectors.graphql
+# inject these rules into source.graphql#graphql-directive, before all other patterns (:L), ignoring strings and comments
+injectionSelector: L:meta.embedded.block.directive.graphql -string -comment
+patterns:
+  - include: "#connect-directive"
+repository:
+  # if we encounter a `@connect` directive, move forward into this set of rules, otherwise it will just follow the default graphql rules
```
phryneas: I added a good bunch of comments to this file, please let me know if they are helpful or if I could make something more clear here.

.vscode/settings.json
```diff
@@ -12,5 +14,8 @@
     "node_modules": false
   },
   "typescript.tsdk": "node_modules/typescript/lib",
-  "debug.node.autoAttach": "on"
+  "debug.node.autoAttach": "on",
+  "yaml.schemas": {
+    "https://raw.githubusercontent.com/martinring/tmlanguage/master/tmlanguage.json": "/syntaxes/*.yaml"
```
phryneas, line 19: I found this JSONSchema which *somehow* adds a tiny little bit of documentation to the textmate format.

src/language-server/__tests__/fixtures/documents/functionCall.ts.snap
```diff
@@ -413,11 +413,11 @@
...
 >(
 #^ source.ts meta.brace.round.ts
->  `query { test }`
+>  `query Q11 { test }`
 #^^ source.ts
 #  ^ source.ts string.template.ts punctuation.definition.string.template.begin.ts
-#   ^^^^^^^^^^^^^^ source.ts string.template.ts
-#                 ^ source.ts string.template.ts punctuation.definition.string.template.end.ts
+#   ^^^^^^^^^^^^^^^^^^ source.ts string.template.ts
+#                     ^ source.ts string.template.ts punctuation.definition.string.template.end.ts
```
phryneas, line 420: Unrelated, it seems I haven't updated some test snapshots in a while.

phryneas:
I think this is in a good state right now, so I'll wait a bit for feedback from testing before we can hopefully get this in next week or so :)

dylan-apollo:
@phryneas this language is going to show up in a few more places now, I'm just wondering how hard it will be to extend syntax highlighting to these as well?

1. In both `@source` and `@connect`, the `http.headers` value is a list of objects or a single object. If an object it has a `value` property (string), which contains `{ }`, then the content within those `{ }` is this same language.
2. Very similar to 1, but _only_ for `@connect`, the string of `http.GET`, `http.POST`, `http.PUT`, `http.PATCH`, and `http.DELETE` can contain expressions within `{ }`.

I believe @BoD said we can use semantic tokens (or whatever they're called) via LSP for JetBrains now, so maybe this is complex enough that we should look into that route?

phryneas:
> @phryneas this language is going to show up in a few more places now, I'm just wondering how hard it will be to extend syntax highlighting to these as well?
> 
>     1. In both `@source` and `@connect`, the `http.headers` value is a list of objects or a single object. If an object it has a `value` property (string), which contains `{ }`, then the content within those `{ }` is this same language.
> 
>     2. Very similar to 1, but _only_ for `@connect`, the string of `http.GET`, `http.POST`, `http.PUT`, `http.PATCH`, and `http.DELETE` can contain expressions within `{ }`.
> 

That would be a duplication of most of `syntaxes/graphql.connectors.json`, with some adjustments. Probably not too complex.

> 
> I believe @BoD said we can use semantic tokens (or whatever they're called) via LSP for JetBrains now, so maybe this is complex enough that we should look into that route?

Semantic tokens won't work for the documentation, and they only work once rover is installed & active, while this approach would always work.
Usually, semantic tokens are used to refine the highlighting of a textmate grammar, not to completely replace it, as afaik it can also be a lot slower and might lag behind while typing. 
In the end, it could also be very difficult to get semantic tokens working for "almost right" code with minor syntax errors (e.g. in situations where you haven't finished typing yet), which oftentimes still kinda works in textmate highlighting - but I don't know how forgiving you can make your parser there.

In the end, it's your call to make.

phryneas:
Okay, finalizing the colors. Here are a few impressions:


image


Here you can see that the object syntax (top) is different from the selection syntax (bottom).
This mimicks the colors that GraphQL syntax is using:

image

Another thing to point out is that builtin functions are highlighted differently:

image

The colors look very different here, but that's theme-specific. I'm using `entity.name.function` for a random function name and `keyword.builtin, entity.name.function` for known function names.

sampleWorkspace/connectors-community/.github/ISSUE_TEMPLATE/new-connector.md
```diff
@@ -0,0 +1,10 @@
+---
+name: New Connector
```
dylan-apollo, line 2: Is this intentionally a full clone of that repo? Should we trim out some bits like this one?

dylan-apollo (approved):
I don't know much about the implementation, but I'm thrilled to have syntax highlighting!

My only note stylistically is I feel like the identifier on the left side of a `:` and the value on the right should have their colors swapped. In the screenshot, it makes sense to be that the short form is the same as the property side of objects (in `headers`). But in the long form, it seems like the key side should match the keys in objects and the value side should match the literal value `$("blah")`.

If that doesn't make sense, happy to huddle on it!

Image

Image

sampleWorkspace/connectors-community/.github/ISSUE_TEMPLATE/new-connector.md
phryneas: Yes, it's a subtree as that's the easiest way to keep those in sync.
```
git subtree pull --prefix sampleWorkspace/connectors-community https://github.com/apollographql/connectors-community.git main --squash
```

There are ways of syncing with subfolders of subtrees, but honestly that didn't seem to be worth the added complexity.

phryneas:
Yeah, happy to huddle on that property colorization issue.

From how I've interpreted it, most languages have two approaches and highlight them differently:

JavaScript: 
```js
const value = 1
// object construction (props and values)
const foo = {
  prop: value
}
// object destructuring (props and aliases)
const { 
  prop,
  prop: alias 
} = foo
```

```graphql
query (value: "1000") {
  # objects in arguments (props and values)
  field(arg: {
    prop: $value
  }) {
    # field selections (fields and aliases)
    field
    alias: field
  }
}
```

And in the connectors syntax I tried to apply that similarly:

```connectors
# inside lit: object construction with props and values
$({
  prop: $value
})

# outside lit: field selection with aliases
alias: field
field
```

image

As that doesn't exist in a vacuum, here the GraphQL example from above next to it so you can see how the color choices are similar:

image

And the JS example:

image



PS: interesting to see how GitHub itself doesn't make that distinction in their code highlighting

chaiphet6669-afk (commented):
`````# - @wtgtybhertgeghgtwtg [__](**increment the:

MAJOR version when you make incompatible API changes MINOR version when you add functionality in a backward compatible manner PATCH version when you make backward compatible bug fixes 

**Example:**

_git tag -s 1.0.1 -m "Bump version: 1.0.0 to 1.0.1_

**)````

- [x] - ```. ![0 comman.txt](https://github.com/user-attachments/files/30298655/0.comman.txt)

chaiphet6669-afk (commented):
sampleWorkspace/connectors-community/.vscode/tasks.json

chaiphet6669-afk (commented):
- - [ ] - - [_sampleWorkspace/connectors-community/README.md_ ] -

PR #243, opened 2024-12-11
[Windows, LSP]: Prevent a console window from opening for the Rover LSP process

phryneas:


dylan-apollo (approved):
Worked for me, thanks!

PR #259, opened 2024-12-31
Report project name in warning when no files are found.

Cellule:
Allows to better pinpoint offending project when working with multiple projects.
I couldn't figure out which of my configs were incorrect and I ended up making this change to debug it on my side.
I figured it wouldn't hurt to push this back upstream

image


.vscode/settings.json
```diff
@@ -6,6 +6,11 @@
   "editor.wordWrapColumn": 110,
   "files.trimTrailingWhitespace": true,
   "files.insertFinalNewline": true,
+  "[typescript]": {
+    "editor.codeActionsOnSave": {
+      "source.organizeImports": "never"
```
Cellule, line 13: I noticed you don't "organize imports" in this project.
So I added this setting to override my default setting.
It might be just a "me" problem, but I figured it makes it makes the way of working more explicit in this project and avoids having to comment on PRs: "Please put back the imports the way they were before"

phryneas: Yeah, that's a good call, thank you!

phryneas (approved):
This looks very useful!

Thanks for the PR and sorry for me being late to it - let's get this in!

PR #280, opened 2025-02-19
Fix detection of new Apollo configuration file

pubmodmatt:
Adding a new configuration file, such as `apollo.config.yaml` or `apollo.config.js`, was not being detected by the VS Code extension. This made it very confusing to try to set up the extension, because adding the config file would not cause the extension to start working. This was due to an incorrect glob pattern, which has been fixed.

Also fixes a race condition when detecting the new config file - it was creating the project, then calling `reloadService`, which immediately destroyed and created the project again, resulting in an error when trying to initialize the already disposed project.

### DIFF
--- a/.changeset/many-trainers-think.md
+++ b/.changeset/many-trainers-think.md
@@ -1,0 +1,5 @@
+---
+"vscode-apollo": patch
+---
+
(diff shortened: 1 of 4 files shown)


phryneas (approved):
Good catches, thanks!

PR #282, opened 2025-02-21
Add block strings to GraphQL grammar

pubmodmatt:
The GraphQL TextMate grammar did not have separate handling for [block strings](https://spec.graphql.org/October2021/#sec-String-Value.Block-Strings), which are different from normal string values in that they may contain unescaped characters like quotes. Block strings were matching the `string.quoted.double.graphql` scope, which did not handle these unescaped characters, leading to incorrect results:

Screenshot 2025-02-21 at 12 35 43 PM

A new `string.quoted.triple.block.graphql` scope is defined, which matches everything up to the closing `"""`:

Screenshot 2025-02-21 at 12 36 42 PM

This becomes more important with Apollo Connectors, where the `selection` attribute is often a block string containing mapping selection syntax that may contain special characters.

### DIFF
--- a/syntaxes/graphql.json
+++ b/syntaxes/graphql.json
@@ -485,6 +485,19 @@
         "1": { "name": "constant.language.null.graphql" }
       }
     },
+    "graphql-block-string-value": {
+      "contentName": "string.quoted.triple.block.graphql",
+      "begin": "\\s*+((\"\"\"))",
+      "end": "\\s*+((\"\"\"))",
+      "beginCaptures": {
+        "1": { "name": "string.quoted.triple.block.graphql" },
+        "2": { "name": "punctuation.definition.string.begin.graphql" }
+      },
+      "endCaptures": {
+        "1": { "name": "string.quoted.triple.block.graphql" },
+        "2": { "name": "punctuation.definition.string.end.graphql" }
+      }
+    },
     "graphql-string-value": {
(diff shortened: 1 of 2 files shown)


syntaxes/graphql.json
```diff
@@ -485,6 +485,19 @@
         "1": { "name": "constant.language.null.graphql" }
       }
     },
+    "graphql-block-string-value": {
+      "contentName": "string.quoted.triple.block.graphql",
+      "begin": "\\s*+((\"\"\"))",
+      "end": "\\s*+((\"\"\"))",
```
nicholascioli, line 491: Do we really want to capture the optional whitespace before the `"""` in both the start and end of the block string? Would just capturing `"""` be enough?

nicholascioli: Also, this might be a dumb question but what does `\s*+` mean?

nicholascioli (approved):
LGTM, though I left a small nit on whitespace capturing.

syntaxes/graphql.json
pubmodmatt: `\s*+` is just possessive matching of whitespace. Since the existing `string.quoted.double.graphql` is using possessive whitespace matching already, we need to use it here as well, or it ends up matching the other pattern since possessive whitespace followed by one " also matches whitespace followed by """.

PR #286, opened 2025-03-19
🔧 Fix rover tests

Jephuff:
Currently, the rover.e2e.ts tests aren't running on any platform, this fixes it so they run on everything but windows (windows needs another fix, so I will enable those tests in that PR)
- Fix the binary path to be the preview that replaced the alpha
- Stop checking for profile in rover.e2e.ts at all, it's not a dependency
- Fix the supergraph.yaml, an empty file is preventing the extension from launching
- Update snapshot for the current rover/lsp version

### DIFF
--- a/sampleWorkspace/rover/apollo.config.yaml
+++ b/sampleWorkspace/rover/apollo.config.yaml
@@ -1,3 +1,3 @@
 rover:
   profile: VSCode-E2E
-  bin: ../../node_modules/@apollo/rover/binary/rover-0.27.0-alpha.0
+  bin: ../../node_modules/@apollo/rover/binary/rover-0.27.0-preview.0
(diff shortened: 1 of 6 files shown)


sampleWorkspace/rover/apollo.config.yaml
```diff
@@ -1,3 +1,3 @@
 rover:
   profile: VSCode-E2E
-  bin: ../../node_modules/@apollo/rover/binary/rover-0.27.0-alpha.0
+  bin: ../../node_modules/@apollo/rover/binary/rover-0.27.0-preview.0
```
pubmodmatt: This seems to be testing with the minimum supported version [defined here](https://github.com/apollographql/vscode-graphql/blob/main/src/language-server/config/config.ts#L398-L400) - does that also need to be updated? Should this be moved up to the release version instead of preview?

Jephuff: Sure can! I was just updating to the installed version, but it definitely should be swapped for the real one 👍

pubmodmatt (approved):

